PR #2, opened 2019-12-25
Support for different types of drilldowns and drilldowns with filters

artkzlv:
**Url example for drilldown with filters:** 
`screen/screenName/view/viewName?filters={"someBcName":"name.contains=11%26status.equalsOneOf=[\\"New\\", \\"Fixed\\"]", "someBcName2": "someField.contains=value"}`
Url parameter "filters" is valid JSON object with bcNames and filters. 
Responsibility for escaping special characters ("%26", "\", etc.) in JSON on whoever writes JSON.

### DIFF
--- a/src/epics/router.ts
+++ b/src/epics/router.ts
@@ -10,6 +10,11 @@ import {ObjectMap} from '../interfaces/objectMap'
 import {makeRelativeUrl, parseBcCursors} from '../utils/history'
 import {buildBcUrl} from '../utils/strings'
 import {DrillDownType, RouteType} from '../interfaces/router'
+import qs from 'query-string'
+import {BcFilter, FilterType} from '../interfaces/filters'
+import {defaultParseLocation} from '../Provider'
+import {shallowCompare} from '../utils/redux'
+import {parsePath} from 'history'
 
 /**
  * Эпик смены текущего маршрута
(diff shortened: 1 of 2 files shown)


src/epics/router.ts
```diff
@@ -188,6 +229,53 @@ const handleRouter: Epic = (action$, store) => action$.ofType(types.handleRouter
...
+                fieldName,
+                type: type as FilterType,
+                value
+            })
+        }
+    })
+    return result.length ? result : null
+}
+
+function isManyValueFilterType(type: string) {
```
Dergash: It's possible to inline that check as it's pretty basic and not reused anywhere

src/epics/router.ts
```diff
@@ -188,6 +229,53 @@ const handleRouter: Epic = (action$, store) => action$.ofType(types.handleRouter
...
+    return (
+        type === FilterType.containsOneOf ||
+        type === FilterType.equalsOneOf
+    )
+}
+
+function locationChanged(pathname: string, currentScreenName: string, currentViewName: string)  {
```
Dergash: [There is a function](https://github.com/tesler-platform/tesler-ui/blob/a0ac06237d57cfa359fdb26ffb7954b8b56b62e9/src/Provider.tsx#L123) that is used for splitting into the `Route` object. This function could be overridden from client side as long as contract have been followed, i.e. it does not necessary will work with this specific splitting algorithm.

`locationChanged` should be called with two `Route` arguments instead

src/epics/router.ts
```diff
@@ -130,12 +133,47 @@ const changeView: Epic = (action$, store) => action$.ofType(types.selectView)
...
-    // const state = store.getState() as Store
+    const state = store.getState() as Store
     const url = action.payload.url
     switch (action.payload.drillDownType) {
-        case 'inner':
+        case DrillDownType.external:
+            window.location.href = url
+            return Observable.empty()
+        case DrillDownType.externalNew:
+            if (/^[a-z0-9]+:\/\//i.test(url)) {
```
Dergash: Will it work for urls without protocol specified, e.g. "www.github.com"?

src/epics/router.ts
```diff
@@ -130,12 +133,47 @@ const changeView: Epic = (action$, store) => action$.ofType(types.selectView)
 
 const drillDown: Epic = (action$, store) => action$.ofType(types.drillDown)
 .switchMap(action => {
-    // const state = store.getState() as Store
+    const state = store.getState() as Store
```
Dergash: `as Store` is not required anymore as `store.getState` return value now has a correct typing

src/epics/router.ts
```diff
@@ -130,12 +133,47 @@ const changeView: Epic = (action$, store) => action$.ofType(types.selectView)
...
     const url = action.payload.url
     switch (action.payload.drillDownType) {
-        case 'inner':
+        case DrillDownType.external:
+            window.location.href = url
+            return Observable.empty()
```
Dergash: There is a bit of inconsistency with `return Observable.empty()` and `break` operators. We end up at `return Observable.empty()` anyway (line 179) so it probably better to just use `break` for case branches

src/epics/router.ts
artkzlv: "window.open" always opens relative url if it came without a protocol

src/epics/router.ts
artkzlv: fixed

src/epics/router.ts
artkzlv: Fixed, function locationChanged removed

src/epics/router.ts
artkzlv: Fixed

src/epics/router.ts
artkzlv: Fixed

PR #12, opened 2019-12-30
Fixed IE11 problem when table page does`t reset after change view

sergromanov4:
Fixed IE11 problem when table page does`t reset after change view

IE11 does`t support Array.from method. 
https://kangax.github.io/compat-table/es6/#ie11 section "Array static methods"


Dergash:
IE11 doesn't support many things; [we handle it](import 'core-js/es/array') with a [core-js](https://github.com/zloirock/core-js/blob/master/packages/core-js/es/array/from.js) polyfill, so it should not have been a problem.

src/reducers/screen.ts
```diff
@@ -75,10 +75,13 @@ export function screen(state = initialState, action: AnyAction): ScreenState {
         }
         case types.selectView: {
             const newBcs: Record<string, BcMetaState> = {}
-            Array
-                .from(
-                    new Set(action.payload.widgets.map(widget => widget.bcName)) // БК которые есть на вьюхе
-                )
+            const bcNameList = action.payload.widgets.map(widget => {
+                if (bcNameList && bcNameList.find(widget.bcName as any)) {
```
Dergash: widget.bcName Is definetly not `any`, it is a string.  
I suspect typescript is correct on this one, as [Array.prototype.find](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find) takes a callback function as first argument, not the value

PR #17, opened 2020-01-01
Decompose <TableWidget /> and fix row menu showing incorrect operations

Dergash:
We should decompose components as much as possible to avoid situations where our `<TableWidget />`  (or any other widget/component, for that matter) will become overgrow with code and functionality.

This PR is concentrated on extracting row operations floating menu into separate component, and preparing our `<TableWidget />` for unit tests (e.g. remove excessive dependencies from interface)

* `<RowOperations />` now handles floating menu.
  * Instead of a single div outside of table, with position and handler controlled via refs, operations are now displayed in last column, so we could contol visibility with a simple css-selector.
  * This should cover the troubles described in #87 when row menu button becomes stuck on single row or rowMeta not fetching causing wrong operations to show
  * `<TableWidget />` no longer depends on Ant Design components `Menu`, `Dropdown`, `Icon` and `Skeleton`, which makes it easier to extract UI part from that component
  * `<TableWidget />` code is much smaller and easier to understand without row menu logic
  * Button design conflicted with row higlight due to round form and white background, now it looks like native action. Additionally, horizontal `ellipsis` icon was replaced with vertical `more` icon to save a bit of space as we using it for tables.
* `<TableRow />` is a kit-agnostic row wrapper that handles our `row operations` feature (i.e. makes additional column and show/hide `<RowOperations />` when necessary) and common row handlers
* Table messages now internationalised
* `<TableWidget />` no longer connects to `router` (handled in epic instead), `bcName` (now directly uses `props.meta.bcName`), `pendingDataItem` (no longer in use), `onDrillDown` (no longer in use) and row menu props `onOperationClick`, `onSelectRow`, `metaInProgress`, they will be removed in 2.0.0

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,18 @@
+# Version 1.5.0
+
+## Features
+
+* Template strings for field titles: Table and Form widgets now support templated title in form of `Title {token:defaultValue}` string, where `token` is replaced with value of the field with key `token` of the currently selected record. `<TemplatedTitle />` component now exported to support this behaviour is custom widgets (#94).
+* `<Pagination />` component now have `onChangePage` callback (#76).
+
+## Fixes
+
+* Hierarchy table should clear active cursor on page change (#76).
+
+## Misc
+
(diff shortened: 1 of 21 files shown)


Dergash:
Resolved in #638

PR #40, opened 2020-01-19
[In progress] New navigation structure

abratashev:


src/interfaces/navigation.ts
```diff
@@ -1,64 +1,44 @@
...
-export interface ViewNavigationCategory {
-    categoryName: string,
-    child: Array<ViewNavigationCategory | ViewNavigationItem>
-}
-
-/**
- * 
- * @param item 
- */
-export function isViewNavigationCategory(item: any): item is ViewNavigationCategory {
```
Dergash: We should keep it until the next major release
```typescript
/**
 * @deprecated TODO: Remove in 2.0.0
 */
export function isViewNavigationCategory(item: any): item is ViewNavigationCategory {
```

abratashev:
Moved here https://github.com/tesler-platform/tesler-ui/pull/40

PR #72, opened 2020-01-31
New TextWidget

anomerova:
Added a new widget for displaying text information with support for markdowns and collapsible section (a new type in the enum of widget types and a new interface)
The [marked](https://marked.js.org/#/README.md#README.md) library is used to display markdowns.


src/components/widgets/TextWidget/TextWidget.tsx
```diff
@@ -0,0 +1,33 @@
+import React from 'react'
+import {WidgetTextMeta} from 'interfaces/widget'
+import {connect} from 'react-redux'
+import {Store} from '../../../interfaces/store'
+import styles from './TextWidget.less'
+
+const parse = require('html-react-parser')
```
Dergash: It is preferrable to use `import` syntax instead of `require`.
`import` syntax is the official JS module system introduced in [ES2015](https://www.ecma-international.org/ecma-262/6.0/#sec-imports).
`require` syntax is unofficial JS module system called "CommonJS" used in NodeJS environment, which nowadays also  has a support for `import` syntax.

Additionaly, you'll need typins for 'marked' library as it is distributed without them:
```sh
yarn add @types/marked -D
```

```tsx
import parse from 'html-react-parser'
import marked from 'marked'
```

npm install @types/marked

src/components/widgets/TextWidget/TextWidget.tsx
```diff
@@ -0,0 +1,33 @@
...
+    meta: WidgetTextMeta
+}
+
+interface TextWidgetProps extends TextWidgetOwnProps {
+    description: string
+}
+
+const TextWidget: React.FunctionComponent<TextWidgetProps> = (props) => {
+    const description = props.description
+    const htmlText = parse(marked(description))
```
Dergash: Will it break if description will be unparsable (non-valid html)? If so then we should check for this and provide a error message instead of page crash.

Dergash:
You should include your component in [tesler-ui/src/components/index.ts](https://github.com/tesler-platform/tesler-ui/blob/9f9b43fe79c9f1c0f33f03dab3b87804a253df57/src/components/index.ts)  
https://github.com/tesler-platform/tesler-ui/blob/9f9b43fe79c9f1c0f33f03dab3b87804a253df57/src/components/index.ts#L19-L21
Otherwise it will not be importable from client app

anomerova:
marked works with passed empty value without errors: the result is an empty div.

Issue #73, opened 2020-01-31
Fix widget hidden type field


AndrewRAM33:
Fix check "hidden" field type on List, DataGrid, Form widgets.
If the field type is "hidden" on these types of widgets, then this field will not be displayed.

Dergash:
Merged https://github.com/tesler-platform/tesler-ui/pull/74

PR #74, opened 2020-01-31
Fix check "hidden" field type on List, DataGrid, Form types widgets. (#73)

AndrewRAM33:
fix/issue-73 Fix check "hidden" field type on List, DataGrid, Form types widgets.

### DIFF
--- a/src/components/widgets/FormWidget/FormWidget.tsx
+++ b/src/components/widgets/FormWidget/FormWidget.tsx
@@ -9,6 +9,7 @@ import Field from '../../Field/Field'
 import {useFlatFormFields} from '../../../hooks/useFlatFormFields'
 import styles from './FormWidget.less'
 import cn from 'classnames'
+import {FieldType} from '../../../interfaces/view'
 
 interface FormWidgetOwnProps {
     meta: WidgetFormMeta,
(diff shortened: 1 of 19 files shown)


src/components/widgets/FormWidget/FormWidget.tsx
```diff
@@ -22,20 +23,24 @@ interface FormWidgetProps extends FormWidgetOwnProps {
 }
 
 export const FormWidget: FunctionComponent<FormWidgetProps> = (props) => {
+    const hiddenKeys: string[] = []
     const flattenWidgetFields = useFlatFormFields<WidgetFormField>(props.meta.fields)
+        .filter((item) => {if (item.type !== FieldType.hidden) { return item } else { hiddenKeys.push(item.key); return undefined}})
```
Dergash: It's a tricky way to filter items, mainly due to them not been actually filtered from an array but replaced with `undefined` values, e.g. `['visible', 'visible', undefined, 'visible', undefined, undefined]` vs. `['visible', 'visible', 'visible]`.
How about something like this?
```tsx
const flattenWidgetFields = useFlatFormFields<WidgetFormField>(props.meta.fields)
.filter(item => {
    const isHidden = item.type === FieldType.hidden
    if (isHidden) {
        hiddenKeys.push(item.key)
    }
    return !isHidden
})
```

AndrewRAM33: yes, this solution is better. Fixed.

PR #75, opened 2020-01-31
Expanded navigation interfaces

abratashev:
UI support of https://github.com/tesler-platform/tesler/issues/18

### DIFF
--- a/src/interfaces/navigation.ts
+++ b/src/interfaces/navigation.ts
@@ -1,28 +1,30 @@
 /**
- * Описание конечного пункта в меню навигации.
+ * Description of the destination in the navigation menu.
  *
- * @param viewName Имя вью.
+ * @param viewName Identifier of view.
  */
 export interface ViewNavigationItem {
-    viewName: string
+    /** TODO identifier will be nullable in 2.0.0 */
+    id?: string
+    viewName: string,
+    hidden?: boolean,
 }
 
 /**
- * TODO
+ * Returns MenuItem if it is ViewNavigationItem
  * 
- * @param item 
+ * @param item to be identified as view
  */
-export function isViewNavigationItem(item: any): item is ViewNavigationItem {
+export function isViewNavigationItem(item: MenuItem): item is ViewNavigationItem {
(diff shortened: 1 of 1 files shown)


abratashev:
First commit:
https://github.com/tesler-platform/tesler-ui/pull/40

src/interfaces/navigation.ts
```diff
@@ -1,64 +1,71 @@
...
- * TODO
- * @param item 
+ * Returns MenuItem if it is ViewNavigationGroup
+ *
+ * @param item to be identified as group
  */
-export function isViewNavigationGroup(item: any): item is ViewNavigationGroup {
-    return !!item && ('title' in item)
+export function isViewNavigationGroup(item: MenuItem): item is ViewNavigationGroup {
+    return !!item && ('child' in item)
```
Dergash: It's a breaking change, we should implement it gradually:
```tsx
    // TODO: remove 'title' check in 2.0.0
    return !!item && ('child' in item || 'title' in item)
```

src/interfaces/navigation.ts
```diff
@@ -1,64 +1,71 @@
...
  *
- * @param title Имя группы.
- * @param child Список категорий или пунктов меню, входящих в группу.
+ * @param title Title of group. Navigation element shows it to user.
+ * @param child Array of navigation elements specified below group(View or inner Group)
  */
 export interface ViewNavigationGroup {
-    title: string
-    child: Array<ViewNavigationItem | ViewNavigationCategory>
+    /** TODO identifier will be nullable in next major */
```
Dergash: It'll be easier to find all planned TODOs if they'll mention target version directly:
```tsx
/** TODO identifier will be nullable in 2.0.0 */
```

src/interfaces/navigation.ts
abratashev: Fixed.

src/interfaces/navigation.ts
abratashev: Fixed.

Issue #87, opened 2020-02-05
Table record menu bugs.


ai-ba:
- Position of the menu button is stuck on one record after selecting a group item. It starts working normally after opening and closing the menu.
- If you open the menu and click on another record, the menu closes, as expected, but when you try to open the menu for clicked record, row-meta is not re-requested and the menu from the previous record opens.

PR #93, opened 2020-02-06
Fix table record menu bugs.

ai-ba:
see #87 

There were two tricky issues that needed workarounds:
\- Dropdown doesn't call menu visibility change handler when we select menu item.
\- On chrome when we select menu item that is located above some record and menu disappears, onmouseenter event that should change menu button location isn't triggered on that record's row.
Table's onmouseenter event keeps working. I use this event to find record's row under cursor and call handler that process row's mouse enter event.


Dergash:
Merged as a temporary fix

Issue #94, opened 2020-02-06
Support for templates in field names (${name}) for Form and Table widgets


artkzlv:
Implement the ability to process templates in field names for Form and Table widgets.
Example of a template in a field name:
'Name is ${fieldName:defaultValue}'

PR #95, opened 2020-02-06
Support for templates in field names (${name}) for Form and Table widgets

artkzlv:
See #94 
Tokens are replaced in the template line with the values ​​of the object fields.
If the value is like a date, then convert it to the format 'DD.MM.YYYY' 

Example:
const item = { color1: 'Green', color2: 'Blue' }
const templatedString = 'Color is ${color1} ${color2:Purple} ${color3:Purple}'
format(templateString, item) // => 'Green Blue Purple'

### DIFF
--- a/src/components/widgets/FormWidget/FormWidget.tsx
+++ b/src/components/widgets/FormWidget/FormWidget.tsx
@@ -10,6 +10,7 @@ import {useFlatFormFields} from '../../../hooks/useFlatFormFields'
 import styles from './FormWidget.less'
 import cn from 'classnames'
 import {FieldType} from '../../../interfaces/view'
+import TemplatedTitle from '../../TemplatedTitle/TemplatedTitle'
 
 interface FormWidgetOwnProps {
     meta: WidgetFormMeta,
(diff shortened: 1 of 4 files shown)


src/components/widgets/FormWidget/FormWidget.tsx
```diff
@@ -51,7 +53,7 @@ export const FormWidget: FunctionComponent<FormWidgetProps> = (props) => {
                         const field = flattenWidgetFields.find(item => item.key === col.fieldKey)
                         const error = (props.missingFields && props.missingFields[field.key])
                             || props.metaErrors && props.metaErrors[field.key]
-                        const fieldLabel = field.type === 'checkbox' ? null : field.label
+                        const fieldLabel = field.type === 'checkbox' ? null : getFieldTitle(field.label, props.data)
```
Dergash: I think we should use a different approach here.  
Instead of introducing dependencies to Form widget, Table widget and all other widgets, which complicates their interface and will create future problems performance-wise, why not introduce a separate component:
```tsx
<TemplatedTitle widgetId={widgetId} title={field.title} container?={CustomControl} />
// ...
const TemplatedWidget = (props) => {
    const wrapper = props.continer && <props.container title={props.templatedTitle} />
    return wrapper || props.templatedTitle
}
// ...
function mapStateToProps(store, ownProps) {
    const widget = store.view.widgets.find(item => item.id === ownProps.widgetId)
    const bc = store.screens.bo.bc[widget.bcName]
    const dataItem = store.data[widget.bcName].find(item => item.id === bc.cursor)
    return {
        templatedTitle: getFieldTitle(ownProps.title, dataItem)
    }
}
export default connect(mapStateToProps)(TemplatedField)
```

src/utils/strings.ts
```diff
@@ -44,3 +46,55 @@ export function buildBcUrl(bcName: string, includeSelf: boolean = false, store?:
     const bcUrl = url.reverse().join('/')
     return bcUrl
 }
+
+// Token format: '${fieldName:defaultValue}'
+const TAG_PLACEHOLDER = /\${(\w*)(:[\wа-яА-ЯёЁ\-,. ]*)?}/g
```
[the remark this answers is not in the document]
artkzlv: Fix the expression

src/components/widgets/FormWidget/FormWidget.tsx
artkzlv: Fixed. Separate component created

Issue #102, opened 2020-02-10
Reset bc page on post-action refreshBC


AndrewRAM33:
In refreshBC post-action, need reset the page to display data correctly (used in sendOperation custom).  CommentWidget incorrectly displays data if you use pagination and add a new item.

PR #103, opened 2020-02-10
New action bcFetchDataPages for fetch data request for specific pages range (#102)

AndrewRAM33:
Added new action bcFetchDataPages for fetch data request for specific pages range. This is used with refreshBC post-action. Allow correctly display data pages in different pagination mode.

### DIFF
--- a/src/epics/screen.ts
+++ b/src/epics/screen.ts
@@ -41,11 +41,23 @@ const processPostInvoke: Epic = (action$, store) => action$.ofType(types.process
         case OperationPostInvokeType.refreshBC: {
             const bo = state.screen.bo
             const postInvoke = action.payload.postInvoke as OperationPostInvokeRefreshBc
+            const postInvokeBC = postInvoke.bc
             const postInvokeBCItem = bo.bc[postInvoke.bc]
-            const widgetName = (action.payload as any).widgetName // TODO: interface should specify widgetName
-            return Observable.of($do.bcFetchDataRequest({
-                bcName: postInvokeBCItem.name, widgetName
-            }))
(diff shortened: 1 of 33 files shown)


src/epics/screen.ts
```diff
@@ -43,9 +43,10 @@ const processPostInvoke: Epic = (action$, store) => action$.ofType(types.process
             const postInvoke = action.payload.postInvoke as OperationPostInvokeRefreshBc
             const postInvokeBCItem = bo.bc[postInvoke.bc]
             const widgetName = (action.payload as any).widgetName // TODO: interface should specify widgetName
-            return Observable.of($do.bcFetchDataRequest({
-                bcName: postInvokeBCItem.name, widgetName
-            }))
+            return Observable.concat(
+                Observable.of($do.bcChangePage({bcName: postInvokeBCItem.name, page: 1})),
```
Dergash: `refreshBc` postInvoke is widely used to refresh data after custom operation which could change displayed record.  
So if the user navigated through several pages of data and then initiated custom operation he will be reverted to page 1 and will lose the record which he interacted with, which is not user-friendly.
At the same time there is a question of infiite scroll paginations, when user fetched several chunks of data with 'Load more' button, will he lose all that and only the first chunk will be displayed?
Additional info required in issue description about the problem this PR is intended to fix.

AndrewRAM33: Undo page reset for bcFetchDataRequest postaction. The bcFetchDataRequest action has been changed, which makes it possible to load pages from the first to the current, with the paginationMode parameter set.

Issue #85, opened 2020-02-04
Filter and sort bugs


AndrewRAM33:
The application has a problem in the operation of the filter and sorting when loading additional pages with data. There is also a problem with re-applying the filter.
**Conditions:**
_1. Set a filter, display a second page with data. The second page will be displayed without filtering.
2. Install the filter. Apply filter again the filter with the same value. All values ​​without a filter are displayed.
3. Set a filter by which there will be no data as a result. Apply filter again. The page will reload.
4. Set a filter, expand the entire list with data, apply sorting. Only the last page data is displayed._

Dergash:
Merged https://github.com/tesler-platform/tesler-ui/pull/86

PR #106, opened 2020-02-16
Undo reset page on sort apply, different action fetch data request for different pagination mode (#85).

AndrewRAM33:
Change sort behavior. When add sort, change page to first is canceled. 
Undo reset page on sort apply,  added different action fetch data request for different pagination mode.

### DIFF
--- a/src/actions/actions.ts
+++ b/src/actions/actions.ts
@@ -586,9 +586,11 @@ export class ActionPayloadTypes {
      * TODO
      * 
      * @param bcName
+     * @param widgetName What widget requires data (widget can only request its own data here)
      */
     bcForceUpdate: {
-        bcName: string
+        bcName: string,
+        widgetName?: string
     } = z
 
     /**
--- a/src/actions/actions.ts
+++ b/src/actions/actions.ts
@@ -131,6 +131,29 @@ export class ActionPayloadTypes {
     } = z
 
     /**
+     * Fetch data request request for specific pages range
+     */
+    bcFetchDataPages: {
+        /**
+         * The business component to fetch data for
+         * @deprecated TODO: 2.0.0 Should be removed in favor of widgetName
+         */
+        bcName: string,
+        /**
+         * Fisrt page to fetch (default is 1)
+         */
+        widgetName: string,
+        /**
+         * What widget requires data (widget can only request its own data here)
+         */
+        from?: number,
+        /**
+         * Last page to fetch (default is current page)
+         */
+        to?: number
+    } = z
+
+    /**
      * Fetch data request for searchable fields
      * 
      * @param bcName The business component to fetch data for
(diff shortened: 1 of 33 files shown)


src/actions/actions.ts
```diff
@@ -121,13 +121,15 @@ export class ActionPayloadTypes {
...
+     * @param paginationMode From first to current page
      */
     bcFetchDataRequest: {
         bcName: string,
         depth?: number,
         widgetName: string,
         ignorePageLimit?: boolean,
-        keepDelta?: boolean
+        keepDelta?: boolean,
+        paginationMode?: boolean
```
Dergash: 1) We should try to keep actions simple and directly connected to user interaction.  
Different use cases equals different actions, as your can see `bcFetchDataRequest` action is already overloaded with totally unintuitive parameters `depth`, `ignorePageLimit` and `keepDelta`, and they'll keep piling up.
I think we need to introduce a new action instead, something like:
```tsx
/**
 * Fetch data request request for specific pages range
 */
bcFetchDataPages: {
    /**
     * The business component to fetch data for
     * @deprecated TODO: 2.0.0 Should be removed in favor of widgetName
     */
    bcName: string,
    /**
     * Fisrt page to fetch (default is 1)
     */
    from?: number,
    /**
     * Last page to fetch (default is current page)
     */
    to?: number
} = z
```
then add this action to epic
```tsx
const bcFetchDataEpic: Epic = (action$, store) => action$.ofType(
    types.bcFetchDataRequest,
    types.bcFetchDataPages,
    // ...other actions
)
```
then 
```tsx
const fetchParams: Record<string, any> = {
    _page: page,
    _limit: limit
}
if (action.type === types.bcFetchDataPages) {
    fetchParams._page = action.payload.from || 1
    fetchParams._limit = (action.payload.to || page - fetchParams._page) * limit
}
```

2) `paginationMode` property name suggests that it'll be either enum or union of some string constants.  
It is difficult to grasp what exactly will happen if this option is active without looking into implementation. So the property shouldn't be boolean or should be named like "allCurrentPages`

src/components/ColumnTitle/ColumnSort.tsx
```diff
@@ -60,12 +60,9 @@ function mapDispatchToProps(dispatch: Dispatch) {
...
             dispatch($do.bcAddSorter({bcName, sorter}))
-            dispatch($do.bcForceUpdate({bcName}))
-            let currentPage = 1
-            while (currentPage < page) {
-                currentPage++
-                dispatch($do.bcLoadMore({bcName}))
-            }
+            page > 1
+                ? dispatch($do.bcForceUpdate({bcName: bcName, paginationMode: true}))
+                : dispatch($do.bcForceUpdate({bcName: bcName}))
```
Dergash: It's simplier to put condition as property value:
```tsx
dispatch($do.bcForceUpdate({ bcName, paginationMode: page > 1 })
```

src/actions/actions.ts
AndrewRAM33: resolved in another PR (#102)

src/components/ColumnTitle/ColumnSort.tsx
AndrewRAM33: Сhanged dispatcher. Another action added.

PR #109, opened 2020-02-18
Added PostInvokeConfirm operation

sergromanov4:
Added the possibility of an additional request to the server after user confirmation.

After customAction is executed, the server response must contain the parameter:

```
preInvoke: {
    type: OperationPreInvokeType //must be of type 'confirm' 
    message: string  // message for user
}
```
After sending a customAction that will return this parameter, the user will have a modal window. If user confirm this action POST request will be sent to the server with parameter 
```
_confirm = confirmOperation.type
```
The server response may contain new Data parameters and postInvoke actions that will also be executed

### DIFF
--- a/src/epics/screen.ts
+++ b/src/epics/screen.ts
@@ -7,7 +7,8 @@ import {
     OperationPostInvokeDrillDown,
     OperationPostInvokeShowMessage,
     OperationPostInvokeDownloadFile,
-    OperationPostInvokeDownloadFileByUrl
+    OperationPostInvokeDownloadFileByUrl,
+    OperationPreInvokeType
 } from '../interfaces/operation'
 import {ObjectMap} from '../interfaces/objectMap'
 import {historyObj} from '../reducers/router'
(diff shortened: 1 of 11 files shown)


src/interfaces/operation.ts
```diff
@@ -268,3 +269,26 @@ export interface OperationErrorEntity {
...
+}
+
+/**
+ * Action, which will be perform after user confirm
+ *
+ * @param type confirm type
+ * @param confirmText text for comfirm modal
+ * @param confirmTitle title for confirm modal
+ */
+export interface OperationPostInvokeConfirm {
```
Dergash: As we've discussed before, those interfaces are literally just a duplicate for existing preinvoke interfaces
https://github.com/tesler-platform/tesler-ui/blob/1b6279833905686c3c0e580b8c6414374336ed0e/src/interfaces/operation.ts#L83-L86

https://github.com/tesler-platform/tesler-ui/blob/1b6279833905686c3c0e580b8c6414374336ed0e/src/interfaces/operation.ts#L91-L107


They should be expanded if necessary and reused, because as we determine earlier preInvokes and postInvoke confirms are the same thing essentially, they just trigger differently (former by a field in operation rowMeta, latter by operation response postInvoke).

src/epics/screen.ts
```diff
@@ -97,8 +99,39 @@ const downloadFileByUrl: Epic = (action$, store) => action$.ofType(types.downloa
...
+                        operationType,
+                        widgetName,
+                        postInvokeConfirm,
+                        isConfirm: true
+                    }))
+                }
+            })
+            return Observable.empty()
+        default:
+            console.warn(`Операция ${action.payload.postInvokeConfirm.type} не поддерживается`)
```
xrav: All messages (including console messages) should be internationalised.

Issue #111, opened 2020-02-20
Support for drilldowns with sorting, reset all filters at once


artkzlv:
Url example for drilldown with sorters:
screen/screenName/view/viewName?sorters={"someBcName":"_sort.0.asc=someField1%26_sort.0.desc=26someField2"}'

"_sort.{order}.{direction}={fieldKey}":
@param fieldKey Sort by field
@param order Priority of this specific sorter
@param direction "asc" or "desc"
Responsibility for escaping special characters ("%26", "", etc.) in JSON on whoever writes JSON.

PR #112, opened 2020-02-20
Support for drilldowns with sorting, reset all filters at once

artkzlv:
See #111

### DIFF
--- a/src/components/ColumnTitle/ColumnFilter.tsx
+++ b/src/components/ColumnTitle/ColumnFilter.tsx
@@ -32,6 +32,20 @@ export const ColumnFilter: FunctionComponent<ColumnFilterProps> = (props) => {
     const {t} = useTranslation()
     const filter = props.filter
     const [value, setValue] = React.useState(filter ? filter.value : null)
+    const [visible, setVisible] = React.useState(false)
+
+    React.useEffect(
+        () => {
+            if (!filter) {
+                setValue(null)
+            }
+        },
+        [filter]
+    )
+
+    const handleVisibleChange = () => {
+        setVisible((prevVisibleState) => !prevVisibleState)
+    }
 
     const handleInputValue = (e: React.ChangeEvent<HTMLInputElement>) => {
         setValue(e.target.value || null)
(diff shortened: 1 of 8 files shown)


src/components/widgets/TableWidget/TableWidget.tsx
```diff
@@ -348,12 +351,19 @@ export const TableWidget: FunctionComponent<TableWidgetProps> = (props) => {
...
     return <div
         className={styles.tableContainer}
         ref={tableContainerRef}
     >
         { props.limitBySelf &&
-            <ActionLink onClick={handleShowAll}>Показать остальные записи</ActionLink>
+            <ActionLink onClick={handleShowAll}> Показать остальные записи </ActionLink>
+        }
+        { props.filters && !!props.filters.length &&
+            <ActionLink onClick={handleRemoveFilters}> Сбросить все фильтры </ActionLink>
```
Dergash:
```tsx
import {useTranslation} from 'react-i18next'
// ...
const {t} = useTranslation()
// ...
<ActionLink onClick={handleRemoveFilters}>
    {t('Clear all filters')}
</ActionLink>
```
Translations assets:
[src/assets/i18n/en.json](https://github.com/tesler-platform/tesler-ui/blob/master/src/assets/i18n/en.json)
[src/assets/i18n/ru.json](https://github.com/tesler-platform/tesler-ui/blob/master/src/assets/i18n/ru.json)

src/components/ColumnTitle/ColumnFilter.tsx
```diff
@@ -89,8 +104,10 @@ export const ColumnFilter: FunctionComponent<ColumnFilterProps> = (props) => {
     return <Popover
         trigger="click"
         content={content}
+        visible={visible}
```
Dergash: You can use [onVisibleChange ](https://ant.design/components/tooltip/#API) property of `<Popover />` to handle the situation when it's not closed on outside click

src/components/widgets/TableWidget/TableWidget.tsx
artkzlv: Fixed

src/components/ColumnTitle/ColumnFilter.tsx
artkzlv: Fixed

src/epics/router.ts
```diff
@@ -270,6 +285,22 @@ function parseFiltersFromString(defaultFilters: string) {
     return result.length ? result : null
 }
 
+/**
+ * Function for parsing sorters from string into BcSorter type
+ */
+function parseSortersFromString(defaultSorters: string) {
```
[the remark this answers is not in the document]
artkzlv: Yes, they are similar, but if 2 or more sortings with same direction will passed, then it will return an array fieldName, not a string

Issue #130, opened 2020-02-28
Broken sort some field List widget type. Filter for a many types.


AndrewRAM33:
Sort buttons are available on a widget List type for each column. However, when using sorting, a 500 error appears on some fields and the field is not sorted.
Extension of column types for which filtering can be applied.

PR #131, opened 2020-02-28
Specified filters for different types of fields. List widget sort exception for columns where it is not necessary.(#130)

AndrewRAM33:
Sort excluded for fields on the List widget, where it cannot be processed.

### DIFF
--- a/src/components/ColumnTitle/ColumnFilter.tsx
+++ b/src/components/ColumnTitle/ColumnFilter.tsx
@@ -32,6 +32,20 @@ export const ColumnFilter: FunctionComponent<ColumnFilterProps> = (props) => {
     const {t} = useTranslation()
     const filter = props.filter
     const [value, setValue] = React.useState(filter ? filter.value : null)
+    const [visible, setVisible] = React.useState(false)
+
+    React.useEffect(
+        () => {
+            if (!filter) {
+                setValue(null)
+            }
+        },
+        [filter]
+    )
+
+    const handleVisibleChange = () => {
+        setVisible((prevVisibleState) => !prevVisibleState)
+    }
 
     const handleInputValue = (e: React.ChangeEvent<HTMLInputElement>) => {
         setValue(e.target.value || null)
@@ -39,6 +53,7 @@ export const ColumnFilter: FunctionComponent<ColumnFilterProps> = (props) => {
 
     const handleApply = (e: FormEvent<HTMLFormElement>) => {
         e.preventDefault()
+        setVisible(false)
         const type = props.widgetMeta.type === FieldType.dictionary
             ? FilterType.equalsOneOf
             : FilterType.contains
@@ -56,7 +71,7 @@ export const ColumnFilter: FunctionComponent<ColumnFilterProps> = (props) => {
 
     const handleCancel = (e: React.MouseEvent<HTMLElement, MouseEvent>) => {
         e.preventDefault()
-        setValue(null)
+        setVisible(false)
         if (props.filter) {
             props.onCancel(props.bcName, filter)
         }
@@ -89,6 +104,8 @@ export const ColumnFilter: FunctionComponent<ColumnFilterProps> = (props) => {
     return <Popover
         trigger="click"
         content={content}
+        visible={visible}
+        onVisibleChange={handleVisibleChange}
     >
         <div
             className={cn(styles.icon, { [styles.active]: !!filter })}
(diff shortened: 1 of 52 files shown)


src/components/ColumnTitle/ColumnTitle.tsx
```diff
@@ -24,12 +25,22 @@ export const ColumnTitle: FunctionComponent<ColumnTitle> = (props) => {
         return <div>{title}</div>
     }
 
-    const filterable = props.rowMeta.filterable
-    const sort = <ColumnSort
+    const noSortable = props.widgetMeta.type === FieldType.multivalue
```
Dergash: It's more readable to write this type of check with the help of Array.prototype.include:
```tsx
const noSortable = [
    FieldType.multivalue,
    FieldType.multifield,
    // ...
    FieldType.hint
].includes(props.widgetMeta.type)
```

src/components/ColumnTitle/ColumnTitle.tsx
```diff
@@ -24,12 +25,22 @@ export const ColumnTitle: FunctionComponent<ColumnTitle> = (props) => {
...
+        || props.widgetMeta.type === FieldType.multifield
+        || props.widgetMeta.type === FieldType.hidden
+        || props.widgetMeta.type === FieldType.fileUpload
+        || props.widgetMeta.type === FieldType.checkbox
```
Dergash: I'm not sure about this one. It seems logical to allow sorting boolean values, so if you have for example github issues with "closed" boolean flag you can sort the issues to show open issues first and then closed.

src/components/ColumnTitle/ColumnTitle.tsx
```diff
@@ -24,12 +25,22 @@ export const ColumnTitle: FunctionComponent<ColumnTitle> = (props) => {
...
+        || props.widgetMeta.type === FieldType.pickList
```
Dergash: I think Picklists also should be sortable. They are major part of application, so if referencing some user with picklist then you'll definetly will want to sort them by they displayed value.  
We do have mapping between picklist value as id and as a displayed string so it should be possible, perhaps with a bit of additional logic, to allow sorting picklists

src/components/ColumnTitle/ColumnTitle.tsx
AndrewRAM33: performed

src/components/ColumnTitle/ColumnTitle.tsx
AndrewRAM33: Yes, that's right. Fixed.

src/components/ColumnTitle/ColumnTitle.tsx
AndrewRAM33: Yes, that's right. Fixed.

src/components/ColumnTitle/ColumnFilter.tsx
```diff
@@ -48,15 +48,47 @@ export const ColumnFilter: FunctionComponent<ColumnFilterProps> = (props) => {
     }
 
     const handleInputValue = (e: React.ChangeEvent<HTMLInputElement>) => {
+        const textValue = e.target.value.substr(0, 100)
+        setValue(textValue || null)
+    }
+
+    const handleDateValue = (date: any, dateString: string) => {
```
Dergash:
```tsx
import {Moment} from 'moment'
// ...
const handleDateValue = (date: Moment, dateString: string) => {
```

src/components/ColumnTitle/ColumnFilter.tsx
```diff
@@ -108,19 +165,19 @@ export const ColumnFilter: FunctionComponent<ColumnFilterProps> = (props) => {
         onVisibleChange={handleVisibleChange}
     >
         <div
-            className={cn(styles.icon, { [styles.active]: !!filter })}
-            dangerouslySetInnerHTML={{ __html: filterIcon }}
+            className={cn(styles.icon, {[styles.active]: !!filter})}
```
Dergash: I believe we follow the rule about spacing on object literals https://eslint.org/docs/rules/object-curly-spacing, so this changes unnecessary. I reconfigure our eslint config to enforce this rule later.

src/components/ColumnTitle/ColumnFilter.tsx
AndrewRAM33: resolve

src/components/ColumnTitle/ColumnFilter.tsx
AndrewRAM33: resolve

Issue #135, opened 2020-03-03
Pass BC name to custom fields.


ai-ba:
The custom field can get the BC name only from the widget meta, which is why when working with the hierarchy with different BCs, the field does not know which BC to use.

PR #136, opened 2020-03-03
Pass hierarchy depth level to custom fields (#135)

ai-ba:
See #135

### DIFF
--- a/src/components/Field/Field.tsx
+++ b/src/components/Field/Field.tsx
@@ -41,6 +41,7 @@ interface FieldOwnProps {
     forceFocus?: boolean,
     forcedValue?: DataValue,
     historyMode?: boolean,
+    customProps?: Record<string, any>,
 }
 
 interface FieldProps extends FieldOwnProps {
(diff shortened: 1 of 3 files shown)


src/components/HierarchyTable/HierarchyTable.tsx
```diff
@@ -232,6 +232,7 @@ export const HierarchyTable: FunctionComponent<HierarchyTableProps> = (props) =>
                         widgetName={props.meta.name}
                         widgetFieldMeta={item}
                         readonly
+                        customProps={{hierarchyDepth: indentLevel + 1}}
```
Dergash: This should memoized or `<Field />` will be recreated every render

src/components/Field/Field.tsx
```diff
@@ -29,6 +29,11 @@ import {CustomizationContext} from '../../components/View/View'
 import {InteractiveInput} from '../../components/ui/InteractiveInput/InteractiveInput'
 import HistoryField from '../../components/ui/HistoryField/HistoryField'
 
+export interface FieldCustomProps {
+    hierarchyDepth?: number,
```
Dergash: This way `Field` still knows about Hierarchy. Custom properties should be of any type and the actual typing for the specific props should be declared in the component that utilizes those props

src/components/HierarchyTable/HierarchyTable.tsx
ai-ba: Fixed.

src/components/Field/Field.tsx
ai-ba: Fixed.

Issue #152, opened 2020-03-24
Editable table is needed


DRITE:
- Cells with no borders, no backlight
- Adding new records to the end of table

Consider for future:
Display ordering by drag-n-drop

PR #164, opened 2020-03-27
Some prepares for EditableTable widget (#152)

DRITE:


src/interfaces/widget.ts
```diff
@@ -166,6 +166,7 @@ export interface WidgetOptions {
 }
 
 export interface WidgetMeta {
+    id?: string,
```
Dergash: There is no `id` property in widget meta description.  
There was once, but we deprecated it in favor of `name` property due to `id` being autogenerated on executing sql data sets of widgets, and so was lacking consistency.

src/components/ui/TextArea/TextArea.tsx
```diff
@@ -135,7 +135,7 @@ const TextArea: React.FunctionComponent<TextAreaProps> = (props) => {
         return <Input.TextArea
             key={key}
             defaultValue={defaultValue}
-            autosize={{
+            autoSize={{
```
Dergash: I think we should keep it `autosize` for now, there is a definite possibility that this will break.

Last 3.*.* version (3.26.13) did indeed deprecated `autosize`, but support both form:
https://github.com/ant-design/ant-design/blob/1c302163afcf2cb97fcb153df9be807fb5200515/components/input/TextArea.tsx#L13

At the same time, previous 3.*.* versions (e.g. 3.18.2) supports only `autosize`:
https://github.com/ant-design/ant-design/blob/6e7c5a95d20cc823417542df3b70272bf708c54c/components/input/TextArea.tsx#L8

As Tesler UI do not have its own instance of antd, client application antd is used, and if they fail to migrate (due to this being minor update, or their own fork) then this property will not work on their side

PR #176, opened 2020-04-05
Add Tag section for AssocListPopup, MVG filter as AssocListPopup

AndrewRAM33:
Added using AssocListPopup widget in popup filters for MVG fields.
Added tag section on AssocListPopup, which displays the selected records and makes it possible to exclude from the selection.

### DIFF
--- a/src/components/ColumnTitle/ColumnFilter.tsx
+++ b/src/components/ColumnTitle/ColumnFilter.tsx
@@ -2,7 +2,7 @@ import React, {FormEvent, FunctionComponent} from 'react'
 import {Popover, Checkbox, Input, Button, Form, Icon, DatePicker} from 'antd'
 import {connect} from 'react-redux'
 import {RowMetaField} from '../../interfaces/rowMeta'
-import {WidgetListField} from '../../interfaces/widget'
+import {MultivalueFieldMeta, WidgetField, WidgetListField, WidgetMeta} from '../../interfaces/widget'
 import styles from './ColumnFilter.less'
 import {$do} from '../../actions/actions'
 import {Store} from '../../interfaces/store'
@@ -25,8 +25,10 @@ export interface ColumnFilterOwnProps {
 export interface ColumnFilterProps extends ColumnFilterOwnProps {
     bcName: string,
     filter: BcFilter,
+    widget: WidgetMeta,
     onApply: (bcName: string, filter: BcFilter) => void,
-    onCancel: (bcName: string, filter: BcFilter) => void
+    onCancel: (bcName: string, filter: BcFilter) => void,
+    onMultivalueAssocOpen: (bcName: string, calleeBCName: string, assocValueKey: string, associateFieldKey: string) => void,
 }
 
 export const ColumnFilter: FunctionComponent<ColumnFilterProps> = (props) => {
@@ -53,6 +55,11 @@ export const ColumnFilter: FunctionComponent<ColumnFilterProps> = (props) => {
 
     const handleDateValue = (date: Moment, dateString: string) => {
         setValue(dateString || null)
+    }
+
+    const handlePopup = (bcName: string, calleeBCName: string, assocValueKey: string, associateFieldKey: string) => {
+        props.onMultivalueAssocOpen(bcName, calleeBCName, assocValueKey, associateFieldKey)
+        handleVisibleChange()
     }
 
     const handleCheckboxValue = (e: CheckboxChangeEvent) => {
@@ -111,8 +118,7 @@ export const ColumnFilter: FunctionComponent<ColumnFilterProps> = (props) => {
     let columnFilterPopover
     switch (props.widgetMeta.type) {
         case (FieldType.dictionary) :
-        case (FieldType.pickList) :
-        case (FieldType.multivalue) : {
+        case (FieldType.pickList) : {
             columnFilterPopover =
                 renderCheckbox(props.widgetMeta.title, value as DataValue[], props.rowMeta.filterValues, setValue)
             break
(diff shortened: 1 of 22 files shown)


src/components/ColumnTitle/ColumnFilter.tsx
```diff
@@ -145,7 +150,10 @@ export const ColumnFilter: FunctionComponent<ColumnFilterProps> = (props) => {
 
     const {t} = useTranslation()
 
-    const content = <div className={styles.content}>
+    const content = props.widgetMeta.type === FieldType.multivalue
+        ? visible &&
+        handlePopup(props.widgetMeta.popupBcName, props.bcName, props.widgetMeta.assocValueKey, props.widgetMeta.associateFieldKey)
```
Dergash, line 155: `handlePopup` is of `(...args) => void` type, so it does not return anything and can not be used as part of logical expression per se. So the only reason it is there is to call side effects - which is a thing that we definetly should try to avoid.
So we have: `visible && handlePopup(args)` will always result in undefined (or void in typescipt environment), as it really is just a `visible && undefined` call, so although it is not clear from the code the content is always `null` for multivalue fields.

If `props.onMultivalueAssocOpen` and `handleVisibleChange` should be called on component mount when field has  multivalue type and visible flag is set, then it is a side effect and should be described with `useEffect` hook:
```tsx
const isMultivalue = props.widgetMeta.type === FieldType.multivalue
const autoOpen = React.useEffect(() => {
    if (isMultivalue  && visible) {
        handlePopup()
    }
}, [dependencies])
// ...
const content = isMultivalue ? undefined : // ...
```

After that you also will not need this `content={content || null}`

src/components/widgets/AssocListPopup/AssocListPopup.tsx
```diff
@@ -56,16 +70,79 @@ export const AssocListPopup: FunctionComponent<IAssocListProps & IAssocListActio
...
+            pendingCurrentData[key].closable = true
+            pendingData.push(pendingCurrentData[key])
+        }
+    }
+
+    // Tag values limit
+    const tagLimit = 5
+    const tagBackgroundCount: number = pendingData.length - tagLimit
+    if (pendingData.length > tagLimit) {
+        pendingData.length = tagLimit
```
Dergash, line 113: Please don't use this pattern to trim the array. It will not lead to a problems here, but if something change with how we fill `pendingData` in the code above (pass it from parent or store for example), we have a very high chances to break something as setting `length` works by reference and may modify some external array.
Prefer `.slice`:
```tsx
const visiblePendingData = pendingData.length > tagLimit
    ? [ ...pendingData.slice(0, tagLimit), { id: '9' } ]
      : pendingData
```

src/components/widgets/AssocListPopup/AssocListPopup.tsx
```diff
@@ -102,10 +179,17 @@ export const AssocListPopup: FunctionComponent<IAssocListProps & IAssocListActio
...
+    const pendingDataChanges = store.view.pendingDataChanges
+    const isFilter = store.view.popupData.isFilter
+    const calleeBCName = store.view.popupData.calleeBCName
     return {
         showed: store.view.popupData.bcName === ownProps.widget.bcName,
         assocValueKey: store.view.popupData.assocValueKey,
-        bcLoading: bc?.loading
+        associateFieldKey: store.view.popupData.associateFieldKey,
+        bcLoading: bc?.loading,
+        pendingDataChanges: pendingDataChanges,
```
Dergash, line 190: It's not clear why you need a separate variable.
```tsx
const pendingDataChanges = store.view.pendingDataChanges
return {
    // ...
    pendingDataChanges
}
```
or
```tsx
return {
    // ...
    pendingDataChanges: store.view.pendingDataChanges
}

File: src/components/widgets/AssocListPopup/AssocListPopup.tsx [hunk 4]
```diff
@@ -56,16 +70,79 @@ export const AssocListPopup: FunctionComponent<IAssocListProps & IAssocListActio
...
+
     const cancelData = React.useCallback(() => {
         onCancel()
         onClose()
     }, [onCancel, onClose])
 
+    const handleDeleteTag = React.useCallback((val: DataItem) => {
+        onDeleteTag(props.widget.bcName, val)
+    },
+        [props.onDeleteTag]
```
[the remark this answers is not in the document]
AndrewRAM33: fixed

src/components/widgets/AssocListPopup/AssocListPopup.tsx
```diff
@@ -56,16 +70,79 @@ export const AssocListPopup: FunctionComponent<IAssocListProps & IAssocListActio
...
+    for (const key in pendingCurrentData) {
+        if (pendingCurrentData[key]._associate === true) {
+            pendingCurrentData[key].closable = true
+            pendingData.push(pendingCurrentData[key])
+        }
+    }
+
+    // Tag values limit
+    const tagLimit = 5
+    const tagBackgroundCount: number = pendingData.length - tagLimit
```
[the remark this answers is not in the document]
AndrewRAM33: fixed

src/components/widgets/AssocListPopup/AssocListPopup.tsx
```diff
@@ -56,16 +70,79 @@ export const AssocListPopup: FunctionComponent<IAssocListProps & IAssocListActio
...
+                id: '99999999999999999999',
+                closable: false,
+                _associate: false,
+                _value: '... ' + tagBackgroundCount
+            }
+        )
+    }
+
+    const tagContent = <div>
+        { (pendingData || []).map(val => {
```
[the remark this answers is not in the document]
AndrewRAM33: fixed

src/components/widgets/AssocListPopup/AssocListPopup.tsx
AndrewRAM33: fixed

src/components/widgets/AssocListPopup/AssocListPopup.tsx
AndrewRAM33: fixed

src/components/ui/Popup/Popup.tsx
```diff
@@ -11,6 +11,7 @@ interface PopupProps {
     onCancelHandler?: () => void,
     size?: 'medium' | 'large',
     children: any,
+    tagContent?: any,
```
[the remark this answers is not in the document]
AndrewRAM33: fixed

src/components/ColumnTitle/ColumnFilter.tsx
AndrewRAM33: fixed with hooks

src/components/widgets/AssocListPopup/AssocListPopup.tsx
```diff
@@ -56,16 +70,79 @@ export const AssocListPopup: FunctionComponent<IAssocListProps & IAssocListActio
...
+                filterValue.push(pendingDataValue[value].id.toString())
+            }
+        }
+        onFilter(props.calleeBCName, {
+            type: FilterType.equalsOneOf,
+            fieldName: props.associateFieldKey,
+            value: filterValue
+        })
+        onClose()
+    }, [onFilter, onClose, props.pendingDataChanges])
```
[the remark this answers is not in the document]
AndrewRAM33: add dep, relocate pendingDataValue

Issue #150, opened 2020-03-24
Required input field return old value


AndrewRAM33:
When clearing the value of a required input field, the old value is returned after the cursor moves to another field.

PR #177, opened 2020-04-05
 Return old value when clear value of a required input field

AndrewRAM33:
**Bug fix, refactoring.** 
The problem with returning the values ​​of the input fields, after clearing them, is that the following code construction lists the types that in the delta (props.pendingValue) can have the value undefined, among which there is no input.
```
const undefinedValuesAllowed = [
        FieldType.pickList,
        FieldType.inlinePickList,
        FieldType.fileUpload,
        FieldType.date,
        FieldType.dateTime,
        FieldType.dateTimeWithSeconds,
        FieldType.checkbox,
        FieldType.dictionary,
        FieldType.multivalue,
        FieldType.number,
        FieldType.money,
        FieldType.percent
    ]
```
The following code block does not provide for when props.pendingValue = '', and the type of the input field is not in the enumeration ('forcedValue' not in props).
```
const value = ('forcedValue' in props)
        ? props.forcedValue
        : (undefinedValuesAllowed.includes(props.widgetFieldMeta.type)
            ? (props.pendingValue !== undefined)
                ? props.pendingValue
                : props.data?.[props.widgetFieldMeta.key]
            : props.pendingValue || props.data?.[props.widgetFieldMeta.key]
        )
```
For any data types present in the delta (props.pendingValue), a value of '' or null is valid (clearing the field). An undefined value cannot be assigned, instead of which props.data?.[props.widgetFieldMeta.key] is assigned. So the case is correct for all types:
```
            (props.pendingValue !== undefined)
                ? props.pendingValue
                : props.data?.[props.widgetFieldMeta.key]
```

### DIFF
--- a/src/components/Field/Field.tsx
+++ b/src/components/Field/Field.tsx
@@ -82,29 +82,12 @@ const simpleDiffSupportedFieldTypes = [
 export const Field: FunctionComponent<FieldProps> = (props) => {
     const [localValue, setLocalValue] = React.useState(null)
     let resultField: React.ReactChild = null
-    // todo: временный фикс для корректной работы с пиклистами
-    const undefinedValuesAllowed = [
-        FieldType.pickList,
-        FieldType.inlinePickList,
-        FieldType.fileUpload,
-        FieldType.date,
-        FieldType.dateTime,
-        FieldType.dateTimeWithSeconds,
-        FieldType.checkbox,
-        FieldType.dictionary,
-        FieldType.multivalue,
-        FieldType.number,
-        FieldType.money,
-        FieldType.percent
-    ]
+
     const value = ('forcedValue' in props)
         ? props.forcedValue
-        : (undefinedValuesAllowed.includes(props.widgetFieldMeta.type)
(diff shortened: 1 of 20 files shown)


src/components/Field/Field.tsx
```diff
@@ -358,7 +358,7 @@ export const Field: FunctionComponent<FieldProps> = (props) => {
                         >
                             <Input
                                 {...commonInputProps}
-                                value={localValue !== null ? localValue : (value ? String(value) : '')}
+                                value={localValue !== null ? localValue : (value  && !props.metaError ? String(value) : '')}
```
Dergash, line 361: 1) It is not clear from the pull request description what the problem was and how it will be fixed, please consider more clear description so the problem is understable from looking at PR without knowing the context
2) It doesn't look right. Check the following gif:
issue-177

The `Country` field does not have a validation error, but it's still resetting its initial value.

AndrewRAM33: refactored

Dergash:
Fixes #150

PR #187, opened 2020-04-09
Location to default screen, when RouteType.default

AndrewRAM33:
It is possible to change the location to default screen. In this case, the ChangeLocation action has `action.payload.location.type: "default"` with empty values `screenName`, and need change location to default screen. Assign the value variable `nextScreenName` default screen name for location to this screen.
The default screen will be the one with the `defaultScreen: true` flag specified. If there are several, then the first of them is used. If there is none, then the first screen of all is used.

### DIFF
--- a/src/epics/router.ts
+++ b/src/epics/router.ts
@@ -36,10 +36,14 @@ const changeLocation: Epic = (action$, store) => action$.ofType(types.changeLoca
         return Observable.of($do.handleRouter(state.router))
     }
 
-    // Если скрины не совпали, то надо просто их перезагрузить и все,
-    // т.к. вьюхи и курсоры зависят от скрина и перестают быть актуальны
+    // Reload screen if nextScreen and currentScreen not equal
+    // With the default route type use the first default screen, if not exist then first screen
     const currentScreenName = state.screen.screenName
-    const nextScreenName = state.router.screenName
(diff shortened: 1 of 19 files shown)


src/epics/router.ts
```diff
@@ -36,10 +36,15 @@ const changeLocation: Epic = (action$, store) => action$.ofType(types.changeLoca
         return Observable.of($do.handleRouter(state.router))
     }
 
-    // Если скрины не совпали, то надо просто их перезагрузить и все,
-    // т.к. вьюхи и курсоры зависят от скрина и перестают быть актуальны
+    // Reload screen if nextScreen and currentScreen not equal
+    // With the default route type use the first default screen, if not exist then first screen
     const currentScreenName = state.screen.screenName
-    const nextScreenName = state.router.screenName
+    const defaultScreenName = state.session.screens.find((screen)=> screen.defaultScreen === true)
```
Dergash, line 42: How about this
```tsx
const defaultScreenName = state.session.sreens.find(screen => screen.defaultScreen)?.name
    || state.session.screens[0]?.name
const nextScreenName = state.router.type === RouteType.default
    ? defaultScreenName
    : state.router.screenName
```

AndrewRAM33: ok, it most shorter

PR #201, opened 2020-04-20
Parse http post 409, 401 status.

AndrewRAM33:
- add case http response status `409 Unauthorized` as logout and redirect to default
- add case http response status `409 Conflict` as system popup error
- refactor ErrorHook


src/utils/api.ts
```diff
@@ -53,38 +55,59 @@ function redirectOccurred(value: AxiosResponse<TeslerResponse>) {
...
+                }
+                break
+            }
+            case 409 : {
+                const notificationMessage = error.response.data.error?.popup?.[0] || ''
+                openButtonWarningNotification(
+                    notificationMessage,
+                    'OK',
+                    0,
+                    () => {Observable.empty()},
```
Dergash, line 91: Observables dObservable.empty is returned in our observables chains to terminate epic reaction to observed action.  
In this case nothing is returned and nothing is observed so it should be just `() => {}`

src/utils/api.ts
```diff
@@ -6,6 +6,8 @@ import {ObjectMap, TeslerResponse} from '../interfaces/objectMap'
 import {$do} from '../actions/actions'
 import {OperationError} from '../interfaces/operation'
 import {BusinessError, ApplicationErrorType} from '../interfaces/view'
+import {openButtonWarningNotification} from './notifications'
+import {historyObj} from '..'
```
Dergash, line 10: I think it would be better if we explicitly declare on which module we are relying here

AndrewRAM33: Ok, changed path to '../reducers/router'

src/utils/api.ts
AndrewRAM33: Fixed

Issue #216, opened 2020-04-29
ModalInvoke `info` and `error` types


AndrewRAM33:
Missing perform `ModalInvoke` window with `info` and `error` type.

PR #217, opened 2020-04-29
Add Modal Invoke window for info and error invoke types (#216)

AndrewRAM33:
Each custom action may be able to perform an additional Invoke operation before and after its execution. Invoke operation displays a modal window for the user. The action will be executed depending on the contents of the modal window and user actions.

The `ModalInvoke` component can be called by the operations` preInvoke` and `postInvoke`. In general, `preInvoke` is an indicator that the next `postAction `implements a condition for further behavior. However, `ModalInvoke` can be used as a reaction to `preInvoke` in simple behavior.
In this case, it is necessary for `ModalInvoke `to be able to process all types of Invoke.

`ModalInvoke` component have been changed, allowing to display modal windows for `info` and `error` invoke types.

* type `info` displays a warning. Сlick on the button is execute custom action.
* type `error` displays an error message. Сlick on the button is **not** execute custom action.
* other types display a window with two buttons. The `OK` button execute custom action, the `CANCEL` button does not.

### DIFF
--- a/src/components/ModalInvoke/ModalInvoke.tsx
+++ b/src/components/ModalInvoke/ModalInvoke.tsx
@@ -5,14 +5,14 @@ import {Store} from '../../interfaces/store'
 import {Modal, Input} from 'antd'
 import {$do} from '../../actions/actions'
 import {useTranslation} from 'react-i18next'
-import {OperationPostInvokeConfirmType, OperationPostInvokeConfirm} from '../../interfaces/operation'
+import {OperationPostInvokeConfirmType, OperationModalInvokeConfirm, OperationPreInvokeType} from '../../interfaces/operation'
 import styles from './ModalInvoke.less'
 
 interface ModalInvokeProps {
     bcName: string,
     operationType: string,
     widgetName: string,
-    confirmOperation: OperationPostInvokeConfirm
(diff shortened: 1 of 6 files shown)


src/components/ModalInvoke/ModalInvoke.tsx
```diff
@@ -29,40 +29,92 @@ const ModalInvoke: React.FunctionComponent<ModalInvokeProps> = (props) => {
...
             case OperationPostInvokeConfirmType.confirmText: {
                 return <div>
-                {props.confirmOperation?.messageContent && <p>{props.confirmOperation?.messageContent}</p>}
+                {props.confirmOperation?.message && <p>{props.confirmOperation?.message}</p>}
                 {<Input value={value} onChange={handleChange}/>}
             </div>
             }
+            case OperationPreInvokeType.info: {
+                return <div>
+                    {props.confirmOperation?.message
```
Dergash, line 43:
```tsx
return <div>
    <p>{ props.confirmationOperation?.message || t('Action has warning') }</p>
</div>
```

AndrewRAM33: *props.confirmOperation?.message
Fixed

PR #223, opened 2020-05-06
Check autoSaveBefore flag directly, dont show popup if false.

AndrewRAM33:
Function check flag `autoSaveBefore` changed. Now the flag value is transmitted directly, which must be processed when displaying the popup of undo changes.

- If `autoSaveBefore = true`, then the record is saved and fields validated before the SendOperation action is performed
- If `autoSaveBefore = false`, the SendOperation action is executed
- If `autoSaveBefore = undefined` validation is performed before the SendOperation action

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,15 @@
+# Version 1.13.0
+
+## Features
+
+* Added optional `bcKey` param to the `sendOperation.associate` action, which will be passed in the `bcName`. It is necessary for identifying the backend which BC to use for popup. (#214)
+* Add Modal Invoke window for info and error invoke types (#217)
+
+## Fixes
+
+* `Field`: Fixed overwriting the default Ant properties of components. If passed property is not defined then it is removed from commonProps and commonInputProps. (#215)
+* Exported ownProps interfaces of all `src/components/ui` components (#219)
+
 # Version 1.12.0
 
 ## Features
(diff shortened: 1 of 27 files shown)


Dergash:
I think we'll have to solve this differently:

1) It is difficult as it is to understand the behavior of `autoSaveBefore` flag, and we do not make it easier by introducing third `undefined` state. It is impossible to understand what `autosaveBefore: undefined` means without knowing the context

2) We do not need the third state, the missed field notification was intentional: even if we want to cancel entity creation, we _need_ to show some kind of a confirmation if user made some changes, because they will be lost. Currently it comes in form of this notification
So when we've encountered that user wants to drop changes without any confirmation at all, I think the way to handle it is to stick with true/false state, and remove the confirmation for `autoSaveBefore: false`. The confirmation then can be reimplemented with `preInvoke` when needed.

3) This change will not cover all cases: only missing required fields, but if the fields values are invalid the custom cancel operation again will fail. I've looked into that and it seems that `autoSaveBefore` on backend side [is not respected for custom operations ](https://github.com/tesler-platform/tesler/blob/c943a2b1dad70c26ce9654bfd906c3c9a832bd4f/tesler-core/src/main/java/io/tesler/core/crudma/impl/inner/InnerCrudmaService.java#L115)

So the plan is to remove notification for `autoSaveBefore: false` and make Tesler API respect this flag for custom operations

Issue #234, opened 2020-05-18
It is necessary to add the ability to customize display dots-menu on the records


artkzlv:
It is necessary to add the ability to optionally display 3 dots (records menu) in the table and ability to add customized columns.

PR #236, opened 2020-05-18
Add the ability to add customized columns in the table

artkzlv:
See #234

The optional `disableDots` property is added in the TableWidget component. When it's true - the table will not show the 3 points (record's menu) on hover.
The optional `controlColumns` property is an array of custom columns and their location.

### DIFF
--- a/src/components/widgets/TableWidget/TableWidget.tsx
+++ b/src/components/widgets/TableWidget/TableWidget.tsx
@@ -1,18 +1,12 @@
-import React, { FunctionComponent } from 'react'
+import React, {FunctionComponent} from 'react'
 import {connect} from 'react-redux'
 import {Dispatch} from 'redux'
-import {
-    Table,
-    Menu,
-    Dropdown,
-    Icon,
-    Skeleton
-} from 'antd'
+import {Dropdown, Icon, Menu, Skeleton, Table} from 'antd'
 import {ColumnProps, TableRowSelection} from 'antd/es/table'
 import ActionLink from '../../ui/ActionLink/ActionLink'
 import {$do} from '../../../actions/actions'
 import {Store} from '../../../interfaces/store'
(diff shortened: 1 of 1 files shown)


src/components/widgets/TableWidget/TableWidget.less
```diff
@@ -82,3 +82,14 @@
         }
     }
 }
+
+.showDotsMenu {
```
Dergash, line 86: We need a broader look for 'Customization' term. It is not enough to add a conditional flag for two different implementations, it is still hardcode and will lead to a bloating of TableWidget component and unintuitive public interface.

What you need is a `controlColumns` property
```tsx
interface ControlColumnProps extends ColumnProps<DataItem> {
    position: 'left' | 'right'
}
interface TableWidgetOwnProps {
    controlColumns?: ControlColumnProps[]
}
// ...
const controlColumnsLeft = props.controlColumns?.filer(item => item.position === 'left')
const controlColumnsRight =  props.controlColumns?.filer(item => item.position === 'right')
const columns = props.meta.fields.map // ...
const resultColumns = [ ...controlColumnsLeft, columns, ...controlColumnsRight ]
```
The default control column is extracted into a separate component:
```tsx
import React from 'react'
import styles  from './defaultControlColumn.less'

export function useDefaultControlColumn(props) {
    const dotColumn: ColumnProps<DataItem> = {
    // ...
    }
}
```
And from the client side:
```tsx
import {TableWidget as TeslerTable, useDefaultControlColumn} from '@tesler-ui/core'

const TableWidget: React.FC = (props) => {
    const controlColumn = useDefaultControlColumn(propsSubset) // do not pass all props, only required subset
    const controlColumns = React.useMemo(() => [controlColumn], [controlColumn])
    return <TeslerTable {...props} controlColumns={controlColumns} />
}
```
This way you will not be caught in situations when you have to further complicate TableWidgets logic due to user design system looks on control differntly: 

tablePattern1

tablePattern2

PR #252, opened 2020-05-27
Autosave pendingDataChanges without pendingValidationFails

AndrewRAM33:
Not execute Autosave pendingDataChanges if there are validation errors pendingValidationFails.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,15 @@
+# Version 1.13.0
+
+## Features
+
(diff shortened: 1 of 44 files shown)


Dergash:
It appears to be the same problem as #257; although this PR will fix the problem with additional save and version mismatch, it will not cover the incorrect notification about pending change and future bugs caused by leaving cell state in the store for forms.

#258 should cover both problems, if not then we'll reopen this.

PR #254, opened 2020-05-27
Exported  ActionPayloadTypes, AnyAction and utilities from actions-utils

DRITE:
see #253


src/index.ts
```diff
@@ -35,6 +36,9 @@ export {
     connect,
     $do,
     types as coreActions,
+    ActionPayloadTypes as CoreActionPayloadTypes,
+    AnyAction as CoreAnyAction,
```
Dergash, line 40: Why are we renaming them? Client application can set the names for themselves during the import

PR #259, opened 2020-06-01
Added the ability to crop string fields and put three dots in the end

sergromanov4:
For Field added a new optional parameter:
`maxValueLength` - this parameter means the maximum allowed number of symbols in the field, after which the end of the word will be found and three dots will be placed

For TableWidget also added optional parameter:
`maxFieldValueLength` - similar parameter for a table to reduce the number of symbols in the table field

### DIFF
--- a/src/components/Field/Field.tsx
+++ b/src/components/Field/Field.tsx
@@ -44,6 +44,7 @@ interface FieldOwnProps {
     forcedValue?: DataValue,
     historyMode?: boolean,
     customProps?: Record<string, any>,
+    maxValueLength?: number
 }
 
 interface FieldProps extends FieldOwnProps {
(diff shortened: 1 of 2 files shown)


Dergash:
Will not be merged
1) `maxValueLength` has no meaning outside of a specific subsets of text fields, i.e. it is no applicable to numbers, files, checkboxes, booleans, multivalues and many others, so there is no reason why `<Field />` should handle it and not the fields that actually need this functionality
2) More than that, it has a potential to break other fields, because the changes are introduced on data level, not visualization. So you've cut the value in `<Field />` component based on its type (`string`) and then passed that value to components like `<Dictionary />` or `<DatePicker />` or `<PickList />`: they all have `string` type of value and they'll receive broken value as a prop, which will not be valid datetime stamp or incorrect value of the dictionary.
3) User will not be able to use browser search when you cut the string programmatically instead of css, so we advise to cut values with `text-overflow: ellipsis` based on column width rather than programmatically based on characters. Also the design decisions based on character numbers will break more easily when applied to languages with other alphabets (compare `ii` and `假借`) and user have increased browser zoom.

If for some reason you can't rely on column width and need to rely on string length then I suggest to reimport a default component, wrap it in HOC with truncate logic and the pass it as a custom field from the client app:
```tsx
// TruncatedInput.tsx
import {Input} from `@tesler-ui/core`

export const TruncatedInput: React.FC = (props) => {
    const valuePreview = props.value.substr(0, 10)
    return <Input {...props} value={valuePreview} />
}
export default TruncatedInput
```
```tsx
// AppLayout.tsx
import {FieldType} from '@tesler-ui/core/interfaces/view'
import TruncatedInput from './TruncatedInput'
// ...
<View
    customFields={{
        [FieldType.input]: TruncatedInput
    }}
/>
```

Issue #213, opened 2020-04-24
Several custom action `associate` on one BC


AndrewRAM33:
Provide the ability to use several custom action `associate` on one BC.
Now, the `sendOperation.associate` action leads to popup BC (`showViewPopup`), with the `bcName` argument. Thus, if there are several `sendOperations.associate` actions, the same popup will be executed.

PR #273, opened 2020-06-04
bcName to API Query Parameters request

AndrewRAM33:
Optional query parameter. Required for the ability to separate by which business component an association operation was performed. (#213)
(tested in version 1.14.2-alpha.0)


src/epics/data.ts
```diff
@@ -518,8 +518,10 @@ const saveAssociationsActive: Epic = (action$, store) => action$.ofType(types.sa
     const pendingChanges = (state.view.pendingDataChanges[
         action.payload.bcNames[0]
     ] || {})
+    let params = {}
+    action.payload.bcNames.forEach(bcName => params = {...params, _bcName: bcName})
```
Dergash, line 522: 1) Only the last bcName will be applied to params
```tsx
let params = {}
['bc1', 'bc2', 'bc3'].forEach(item => params = { ...params, _bcName: item })
console.log(params) // {_bcName: "bc3"}
```
2) You generally do not want to recreate an object inside `forEach` call when you can simply assign the field:
```tsx
action.payload.bcNames.forEach(bcName => {
    params._bcName = bcName
})
```
3) Curly bracers are mandatory for any assignments

AndrewRAM33: fixed

Dergash: It's still strangely iterates over an array to set the value once.
```tsx
const params: GetParamsMap = {}
action.payload.bcNames.forEach(bcName => {params._bcName = bcName})
```
What is happening here? Why you always use only the last bcName if there are multiple of them and if for some reason you do need the last one, why are you not accessing it directly?

src/epics/data.ts
```diff
@@ -518,8 +519,10 @@ const saveAssociationsActive: Epic = (action$, store) => action$.ofType(types.sa
     const pendingChanges = (state.view.pendingDataChanges[
         action.payload.bcNames[0]
     ] || {})
+    const params: GetParamsMap = {}
```
Dergash, line 522: Record<string, any>

src/epics/data.ts
AndrewRAM33: Fixed. Take to param only last bcName from array.
```
const bcNames = ['a','b','c']
const params = bcNames.length ? {_bcName: bcNames.pop()} : {}
console.log(params) // {_bcName: "c"}
```

src/epics/data.ts
AndrewRAM33: fixed

src/epics/data.ts
```diff
@@ -518,8 +518,8 @@ const saveAssociationsActive: Epic = (action$, store) => action$.ofType(types.sa
     const pendingChanges = (state.view.pendingDataChanges[
         action.payload.bcNames[0]
     ] || {})
-
-    return api.associate(state.screen.screenName, bcUrl, Object.values(pendingChanges) as AssociatedItem[])
+    const params: Record<string, any> = action.payload.bcNames.length ? {_bcName: action.payload.bcNames.pop()} : {}
```
Dergash, line 521: `Array.prototype.pop` will modify the source array (action.payload.bcNames) and incorrect array will be forwarder through an action chain (e.g. bcCancelPendingChanges)

AndrewRAM33: fixed as `array[array.length - 1]`

PR #277, opened 2020-06-10
Select all checkbox in FullHierarchyTable

sergromanov4:
Fixed "Select all" checkbox  in FullHierarchyTable.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+# Version 1.14.1
+
+## Fixes
+
+* `limit` calculated incorrectly after 1.14.0 leading to broken pagination ([#264](https://github.com/tesler-platform/tesler-ui/issues/264))
+* `<TableWidget />` does not recalculate columns after 1.14.0 due to missing prop dependency leading to missing filtration and sorting ([#270](https://github.com/tesler-platform/tesler-ui/issues/270))
+
 # Version 1.14.0
 
 ## Features
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@tesler-ui/core",
-  "version": "1.14.0",
+  "version": "1.14.1",
   "main": "tesler-ui-core.js",
   "homepage": "https://tesler.io/",
   "types": "index.d.ts",
--- a/src/components/FullHierarchyTable/FullHierarchyTable.tsx
+++ b/src/components/FullHierarchyTable/FullHierarchyTable.tsx
@@ -33,6 +33,7 @@ interface FullHierarchyTableDispatchProps {
     onSelect: (bcName: string, depth: number, dataItem: AssociatedItem, widgetName: string, assocValueKey: string) => void,
     onDeselectAll: (bcName: string, depthFrom: number) => void,
     onSelectAll: (bcName: string, parentId: string, depth: number, assocValueKey: string, selected: boolean) => void,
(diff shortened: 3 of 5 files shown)


src/actions/actions.ts
```diff
@@ -511,6 +511,21 @@ export class ActionPayloadTypes {
...
+     * TODO
+     * 
+     * @param bcName
+     * @param dataItems
+     * @param assocValueKey
+     * @param selected
+     */
+    changeDataItemsFull: {
+        bcName: string,
+        dataItems: PendingDataItem[],
```
Dergash, line 524: * You do not need to pass the data in payload as it is [already in the store](https://github.com/tesler-platform/tesler-ui/blob/728d8b5ea2bc630e72bcc82462c9b79ae1db8f14/src/components/FullHierarchyTable/FullHierarchyTable.tsx#L230); moreover it will be memory consuming considering that you passing all items of a large table hierarchy.
* You do not need `bcName` as `widgetName` give you more options and `bcName` usually directly derived from widget
* And `assocValueKey` as it is already in the store ([store.view.popupData.assocValueKey](https://github.com/tesler-platform/tesler-ui/blob/728d8b5ea2bc630e72bcc82462c9b79ae1db8f14/src/components/widgets/AssocListPopup/AssocListPopup.tsx#L187))
* `selected` also can be deduced by reducer based on `pendingChanges` and `store.data[bcName]` lengths - if they do not match then you mark all items as selected, if they match then you mark all items as deselected.

src/reducers/view.ts
```diff
@@ -232,6 +232,22 @@ export function view(state = initialState, action: AnyAction, store: Store) {
                 }
             }
         }
+        case types.changeDataItemsFull: {
+            const newPendingChanges = { ...state.pendingDataChanges[action.payload.bcName] }
+            action.payload.dataItems.forEach((dataItem: PendingDataItem) => {
```
Dergash, line 237: You do not need an explicit type for `dataItem` argument

src/actions/actions.ts
```diff
@@ -511,6 +511,21 @@ export class ActionPayloadTypes {
...
 
+    /**
+     * TODO
+     * 
+     * @param bcName
+     * @param dataItems
+     * @param assocValueKey
+     * @param selected
+     */
+    changeDataItemsFull: {
```
Dergash, line 522: This name is not valid; `changeDataItem` means changing data item in any way when this particular actions does not interact with data items at all and only handles additional, association widgets specific fields `_associate` and `_value`

Something like `setAllAssociationsFull` would work, but it appears there is already a similiar action implemented, check it out: [changeChildrenAssociations](https://github.com/tesler-platform/tesler-ui/blob/728d8b5ea2bc630e72bcc82462c9b79ae1db8f14/src/epics/data.ts#L535).
From looking at the code it handles exactly the same thing so this pull request could be collapsed to simply proprerly cover the existing action with JSDoc

src/components/FullHierarchyTable/FullHierarchyTable.test.tsx
```diff
@@ -39,6 +39,7 @@ describe('FullHierarchyTable test', () => {
         onSelect: jest.fn(),
         onDeselectAll: jest.fn(),
         onSelectAll: jest.fn(),
+        onSelectFullTable: jest.fn(),
```
Dergash, line 42: This actually tells you that it is a breaking change; consider unit test a miniature client app, if it breaks, then the client application who import our unconnected `<FullHierarchyTable />` component will also break, meaning `onSelectFullTable` property should be optional

Issue #282, opened 2020-06-13
preinvoke execution before sendOperation action


AndrewRAM33:
Need realize ability to perform pre-action before `sendOperation` action.
`preInvoke` presents a confirmation or notification window that will be displayed before is executed `sendOperation` action.  `SendOperation` will be executed or not depend of type `preInvoke` action and user's actions.

PR #283, opened 2020-06-13
preInvoke Middleware

AndrewRAM33:
preInvoke Middleware (#282)

`preInvoke` presents a confirmation or notification window that will be displayed before is executed `sendOperation` action.  `SendOperation` will be executed or not depend of type `preInvoke` action and user's actions.

Type of preInvoke action:
- `confirm` - window with text and yes / no buttons to confirm execution
- `info` - window with info message, `sendOperation` executed.
- `error` - window with error message, `sendOperation` not executed.

Features:
- Popup bottons custom text

Example preInvoke operation:
```
confirmOperation:
    message: 'Are you sure?'
    type: 'confirm'
```


src/components/widgets/TableWidget/TableWidget.tsx
```diff
@@ -55,7 +55,7 @@ export interface TableWidgetProps extends TableWidgetOwnProps {
     filterGroups: FilterGroup[],
     onDrillDown: (widgetName: string, bcName: string, cursor: string, fieldKey: string) => void,
     onShowAll: (bcName: string, cursor: string, route: Route, widgetName: string) => void,
-    onOperationClick: (bcName: string, operationType: string, widgetName: string) => void,
+    onOperationClick: (bcName: string, operationType: string, widgetName: string, bcKey?: string, preInvoke?: OperationPreInvoke) => void,
```
Dergash, line 58: We should avoid complicating functions contract when possible.  
There are already a `widgetName` and `operationType` arguments, you can use them inside the middleware to find the widget, the operation and get `bcKey` and `preInvoke` from there

AndrewRAM33: fixed. Use from state.view.rowMeta

src/components/ui/Popup/Popup.tsx
```diff
@@ -44,10 +46,10 @@ export const Popup: FunctionComponent<PopupProps> = (props) => {
                     }
                     <div className={styles.actions}>
                         <Button onClick={props.onOkHandler} className={styles.buttonYellow}>
-                            {t('Save')}
+                            {props.OkText ? props.OkText : t('Save')}
```
Dergash, line 49:
```tsx
{props.OkText ?? t('Save')}
```

src/components/ui/Popup/Popup.tsx
```diff
@@ -44,10 +46,10 @@ export const Popup: FunctionComponent<PopupProps> = (props) => {
...
                         </Button>
                         <Button onClick={props.onCancelHandler} className={styles.buttonCancel}>
-                            {t('Cancel')}
+                            {props.cancelText ? props.cancelText : t('Cancel')}
```
Dergash, line 52:
```tsx
{props.cancelText ?? t('Cancel')}
```

src/middlewares/preInvokeMiddleware.tsx
```diff
@@ -0,0 +1,34 @@
...
+import {$do, types} from '../actions/actions'
+import {Store as CoreStore} from '../interfaces/store'
+
+const preInvokeAction = ({getState, dispatch}: MiddlewareAPI<Dispatch<AnyAction>, CoreStore>) => (next: Dispatch) =>
+    (action: AnyAction) => {
+        if (action.type === types.sendOperation) {
+            const bcName = action.payload.bcName
+            const operationType = action.payload.operationType
+            const widgetName = action.payload.widgetName
+            const preInvoke = action.payload?.confirmOperation
```
Dergash, line 15: `confirmOperation` field is deprecated:
https://github.com/tesler-platform/tesler-ui/blob/04f4874f83f6072ad938bdb749cb2d8cd5d03899/src/actions/actions.ts#L339-L342

IDE would show it if not for the missing typings; in some cases typescript can deduce the type based on the condition branch but not always: in this case it couldn't so you'll have to use manual case or write a typeguard to get correct typings for payload.

And with correct typings IDE will show that `confirmOperation` field is deprecated.
The reasoning is that when you need a pre-invoke you can calculate it based on `widgetName` and `operationType`: get the widget, read bcName, use bcName to access current row meta, use operationType to get the operation, read the `preInvoke` from there.  
The `confirmOperation` field is optional and intended for cases when you hardcode confirmation logic from front end.

src/middlewares/preInvokeMiddleware.tsx
```diff
@@ -0,0 +1,34 @@
...
+
+            return preInvoke && !action.payload.confirm
```
Dergash, line 17: You probably do not need `action.payload.confirm` here

AndrewRAM33: It necessary that `preInvokeMiddleware`  not execute when confirm response is received

src/components/ui/Popup/Popup.tsx
AndrewRAM33: fixed

src/components/ui/Popup/Popup.tsx
AndrewRAM33: fixed

src/middlewares/preInvokeMiddleware.tsx
AndrewRAM33: fixed. Get `preInvoke` Operation from state

PR #290, opened 2020-06-15
Basis for the save number of Tablewidget records

sergromanov4:
Basis for storing the number TableWidget records that you can use in your project.

### DIFF
--- a/src/actions/actions.ts
+++ b/src/actions/actions.ts
@@ -369,6 +369,26 @@ export class ActionPayloadTypes {
     } = z
 
     /**
+     * Fetch number of table records
+     * 
+     * @param widgetName What widget requires data
+     */
+    bcFetchRecordCounter: {
+        widgetName: string,
+    } = z
+
+    /**
+     * Fetch number of table records request was succesful
+     * 
+     * @param widgetName Business component that initiated data fetch
+     * @param countRecords number of table records
+     */
+    bcFetchRecordCounterSuccess: {
+        widgetName: string,
+        countRecords: number,
+    } = z
+
+    /**
      * TODO
      * 
      * @param postInvoke
--- a/src/actions/actions.ts
+++ b/src/actions/actions.ts
@@ -369,26 +369,6 @@ export class ActionPayloadTypes {
     } = z
 
     /**
-     * Fetch number of table records
-     * 
-     * @param widgetName What widget requires data
-     */
-    bcFetchRecordCounter: {
-        widgetName: string,
-    } = z
-
-    /**
-     * Fetch number of table records request was succesful
-     * 
-     * @param widgetName Business component that initiated data fetch
-     * @param countRecords number of table records
-     */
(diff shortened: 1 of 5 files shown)


src/actions/actions.ts
```diff
@@ -368,6 +368,26 @@ export class ActionPayloadTypes {
         cursor: string
     } = z
 
+    /**
+     * Fetch number of table records
+     * 
+     * @param bcName Business component that initiated data fetch
+     */
+    bcFetchRecordCounter: {
+        bcName: string,
```
Dergash, line 377: We should prefer `widgetName` instead of `bcName` when possible.  
For example, here we have an action that requires records counting - we request it from some widget, and there are  cases when it's important to know which specific widget was a source of this request, because if there are several widgets sharing the same business component there is really no way to tell them apart. So it's better to pass the widget name in actions payloads and then derive bcName from this widget.

Use bcName only in situations when there is no initiating widget

src/actions/actions.ts
```diff
@@ -368,6 +368,26 @@ export class ActionPayloadTypes {
...
+
+    /**
+     * Fetch number of table records request was succesful
+     * 
+     * @param bcName Business component that initiated data fetch
+     * @param countRecords number of table records
+     */
+    bcFetchRecordCounterSuccess: {
+        bcName: string,
+        countRecords: DataValue,
```
Dergash, line 388: It doesn't seem right.  
`DataValue` is the union of possible types for values of data fetching, while you trying to use it as a generic `Record<string, number>` type for your `{ [bcName: string]: number }` structure.
1) You can't just use the unrelated type based on the basis that structurally it matches you goals; you should introduce your own type
2) `DataValue` actually is not a match, that's why you later exlicitly converting it number
3) You probably need here just a regular `number`, it is strange to have a success action with one bc, but with countRecords being a dictionary of results for several bcs

src/index.ts
```diff
@@ -18,12 +18,14 @@ export * from './hooks'
 
 import {historyObj, changeLocation} from './reducers/router'
 export {
+    applyParams,
```
Dergash, line 21: We probably shouldn't expose this helper:
1) It's really tiny, even counting two nested functions it's only 12 lines of trivial code, it's easy to implement from the client side.
2) It is a generic helper not connected to Tesler functionallity in any way. We have a number of generic helpers and utilities in our code base related to urls, strings, dates, etc. We can't just expose all of them just because client app requires similiar functionallity and we have it implemented. There is a public API of tesler-ui and as soon as something exposed and becomes a part of this public API we have to consider semantic versioning for this helpers, meaning we are no longer free to change their contract or behavior.

Dergash:
It turns out there were no intention to elevate this functionality from project level to the Tesler UI, and as no API call implemented here and likewise no corresponging endpoint exist on Tesler API side, all the code moved to the project level instead.

I'll keep the export of `getFilters` though as this indeed will be required by any projects calling Tesler API endpoints with filtration support.

PR #293, opened 2020-06-17
PickListPopup remove footer buttons and fix title

sergromanov4:
Removed buttons from footer in PickListPopup.
Fixed title in PickListPopup.


src/components/ui/Popup/Popup.tsx
```diff
@@ -16,6 +16,7 @@ export interface PopupProps {
     bcName: string,
     widgetName?: string,
     disablePagination?: boolean,
+    disableFooterButtons?: boolean,
```
Dergash, line 19: We need a broader approach to customization. We should introduce a 'footer: React.ReactNode' property: if this property was passed down to the `<Popup />` component, we render it. If this property was not passed, then we render current default implementation. If the `null` was passed, we render nothing (your case).

This way we will not end up with bloated component API

PR #309, opened 2020-06-29
Fixed bug when preInvoke doesn`t work if you use operation in dots menu

sergromanov4:
Fixed bug when preInvoke doesn`t work if you use operation in dots menu in the TableWidget


Dergash:
I'm going to close this in favor of https://github.com/tesler-platform/tesler-ui/pull/283.  
It is intended behaviour and conscious choice to not explicitly pass `confirmOperation` in `sendAction` payload from `<TableWidget />`.  
The reasoning is that when your `confirmOperation` is operation pre-invoke, you can calculate it from the `widgetName` and `operationType` in `sendOperation` payload.  
So `confirmOperation` property in `sendOperation` payload intended for cases when you want to apply some confirmation manually, not when you already have a preinvoke assigned by Tesler API.

PR #314, opened 2020-07-01
Fixed configuration

DRITE:
There was the mistake in PR #281 

### Reasons
Let's consider reasons as chain of facts:
1. Old version of `tsconfig.json` contained option `"outDir": ""`
2. In PR #281 `ts-loader` has been upgraded to `6.2.2`
3. New version of `ts-loader` according to option `"outDir": ""` generates declaretion files in current directory.
Thats why option `"outDir": "dist"` has been added.
That was the reason of next error:
While compiling `ts-loader` concatinates pathes from option `"outDir": "dist"` of `tsconfig.json` and `output.path` of `webpack.config.js`. Thats why there was another `dist` folder in `dist` folder.

### Fixes
 In this PR `tsconfig.json` was fixed by option `"outDir": "./dist"`. It leads to correct structure in `dist` folder.

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -5,6 +5,8 @@ const rxjsExternals = require('webpack-rxjs-externals')
 const { CleanWebpackPlugin } = require('clean-webpack-plugin')
 const CopyWebpackPlugin = require('copy-webpack-plugin')
 
+const isProduction = process.argv.some(arg => arg === '-p' || arg === '--production' || arg === '--prod')
+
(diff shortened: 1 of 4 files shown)


webpack.config.js
```diff
@@ -26,9 +28,9 @@ module.exports = (env, options) => {
     return  {
         entry: ['./src/index.ts'],
         mode: options.mode || 'development',
+        watch: isProduction,
```
Dergash, line 31: Why is watch enabled for production?

Issue #321, opened 2020-07-03
sendOperation with custom param


AndrewRAM33:
Deprecate `bcKey` in favor of `customParam` `[customParam: string]:any` with custom implementation.

Dergash:
Rejected in favor of #330

PR #322, opened 2020-07-03
Deprecate bcKey in favor of customParam with custom implementation.

AndrewRAM33:
- Deprecate `bcKey` in favor of `customParam` with custom implementation. Now you can add custom parameters to the `sendOperation` action. In this case, _**Tesler-ui**_ implementation will not be used (`sendOperation.associate`, `sendOperation.create`).
- Dont show `<Tag>` in `AssocListPopup` if empty value
(#321)

### DIFF
--- a/src/actions/actions.ts
+++ b/src/actions/actions.ts
@@ -327,6 +327,7 @@ export class ActionPayloadTypes {
      * @param onSuccessAction Any other action
      * @param confirm params for confirm modal
      * @param bcKey key called bk
+     * @param customParam custom any parameter
      */
     sendOperation: {
         bcName: string,
@@ -334,7 +335,9 @@ export class ActionPayloadTypes {
         widgetName: string,
         onSuccessAction?: AnyAction,
         confirm?: string,
-        // TODO: Will not be optional in 2.0.0
+        /**
+         * @deprecated TODO: Remove in 2.0.0
+         */
         bcKey?: string,
         /**
          * @deprecated TODO: Remove in 2.0.0 in favor of sendOperationWithConfirm
--- a/src/actions/actions.ts
+++ b/src/actions/actions.ts
@@ -327,7 +327,6 @@ export class ActionPayloadTypes {
      * @param onSuccessAction Any other action
      * @param confirm params for confirm modal
      * @param bcKey key called bk
-     * @param customParam custom any parameter
      */
     sendOperation: {
         bcName: string,
--- a/src/epics/data.ts
+++ b/src/epics/data.ts
@@ -292,30 +292,32 @@ const bcNewDataEpic: Epic = (action$, store) => action$.ofType(types.sendOperati
 .filter(action => action.payload.operationType === OperationTypeCrud.create)
 .mergeMap((action) => {
     const state = store.getState() as Store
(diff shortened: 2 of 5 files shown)


src/actions/actions.ts
```diff
@@ -327,15 +327,19 @@ export class ActionPayloadTypes {
...
         operationType: OperationTypeCrud | string,
         widgetName: string,
         onSuccessAction?: AnyAction,
         confirm?: string,
-        // TODO: Will not be optional in 2.0.0
+        /**
+         * @deprecated TODO: Remove in 2.0.0 in favor of customParam
+         */
         bcKey?: string,
+        [customParam: string]:any,
```
Dergash, line 342: This will not be merged.

1) When introducing `[someVar: someType]: any` pattern to your type it is important to wrap into into a separate property:
```tsx
interface someInterface {
  // ...
  customProperties: Record<string, any>
}
```
Otherwise you will be open for misspelling mistakes:
```tsx
sendOperation({ onSucessAction })
// no error will be thrown because `onSucessAction` now considered `[customParam: string]`  
```
2) This is an overall incorrect approach to allow core actions have unspecified payload and bind customization logic to the fact of the presence of said payload.  
When client app wants an additional payload parameter it is the client app responsibility to extend core type and dispatch an extended actions instead of the core one:

```tsx
import {ActionPayloadTypes} from '@tesler-ui/core'
export class CustomActionPayloadTypes extends ActionPayloadTypes {
  // ...
  sendOperation: ActionPayloadTypes ['sendOperation'] & {
        customParam: number
    } = z
}
```
And then build observable action creators from this class and then dispatch extended action like this:
```tsx
dispatch($clientDo.sendOperation({
  // core payload
  // ...
  customParam: number
})
```
And then handle this additional parameter in a client side reducer.  
No additional code in Tesler UI is necessary for that.

src/epics/data.ts
```diff
@@ -292,30 +292,31 @@ const bcNewDataEpic: Epic = (action$, store) => action$.ofType(types.sendOperati
...
-            postInvoke
-                ? Observable.of($do.processPostInvoke({ bcName, postInvoke, cursor, widgetName: action.payload.widgetName }))
-                : Observable.empty<never>()
-        )
-    })
-    .catch((error: any) => {
-        console.log(error)
-        return Observable.of($do.bcNewDataFail({ bcName }))
-    })
+    return Object.keys(customParam).length === 0
```
Dergash, line 297: This pattern better handled with early return:
```tsx
if (!Object.keys(customParam).length) {
 return Observable.empty<never>()
]
return api.newBcData(state.screen.screenName, bcUrl, {widgetName})
```
This way change set is more clear.

AndrewRAM33: fixed

src/actions/actions.ts
AndrewRAM33: custom Param moved to client application

AndrewRAM33:
fix review comments

Dergash:
Reverted epic changes due to introducing https://github.com/tesler-platform/tesler-ui/pull/331
Fix a potential problem with regards using zero number in congitional mapping

Issue #374, opened 2020-08-02
Refresh BC after drilldown with filters


AndrewRAM33:
Application want refresh bc, when recived drilldown wih refresh. 

If we have many view with another content and same one of bc, and routing between this views on drilldown with filters on same bc, then bc dont update data and will display old records.

PR #375, opened 2020-08-02
Refresh bc on drilldown with filter

AndrewRAM33:
Refresh bc on drilldown with filter (#374)

After receive filter witn drilldown need refresh filtered bc. This can lead to bugs when routing between views containing the same bc.

### DIFF
--- a/src/epics/router.ts
+++ b/src/epics/router.ts
@@ -26,7 +26,7 @@ const changeLocation: Epic = (action$, store) => action$.ofType(types.changeLoca
 .mergeMap(action => {
     const state = store.getState()
 
-    // пользователь еще не авторизовался
+    // User not logged
     if (!state.session.active) {
         return Observable.empty()
     }
@@ -49,7 +49,7 @@ const changeLocation: Epic = (action$, store) => action$.ofType(types.changeLoca
             ? Observable.of($do.selectScreen({ screen: nextScreen }))
             : Observable.of($do.selectScreenFail({ screenName: nextScreenName }))
     }
-    // Проверить, есть ли расхождение по курсорам БК в урле и в сторе
+    // Check cursor different between store and url
     const currentViewName = state.view.name
     const nextViewName = state.router.viewName
     const nextCursors = parseBcCursors(state.router.bcPath) || {}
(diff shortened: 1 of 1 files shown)


src/epics/router.ts
```diff
@@ -185,12 +185,14 @@ const drillDown: Epic = (action$, store) => action$.ofType(types.drillDown)
...
                                 parsedFilters.forEach((item) => {
                                     store.dispatch($do.bcAddFilter({bcName, filter: item}))
+                                    store.dispatch($do.bcForceUpdate({bcName}))
                                 })
                                 if (!diff.length) {
                                     store.dispatch($do.bcForceUpdate({bcName}))
                                 }
                             } else {
                                 store.dispatch($do.bcRemoveAllFilters({bcName}))
+                                store.dispatch($do.bcForceUpdate({bcName}))
```
Dergash, line 195: * You are calling `bcForceUpdate` in each branch, => it should be one call outside of branches
* You are pushing `bcForceUpdate` for the same BC on each filter, meaning if you filter one business component by five different columns, this bc will fetch data five times and rerenders more than five times.
* overall it doesn't seem like a proper place to fix it. From the #374 I grasp that if we have view1 and view2 with the same widget, than filtration drilldown from view1 to view2 will apply new filters, but will not refetch the data with new filters. It's unlikely as `changeLocation` by itself causes updates (a little more likely if we talking about filtration drilldown onto the same view), but either way I suspect it should be handled the same way as we handle "same view, different cursors" scenario:
https://github.com/tesler-platform/tesler-ui/blob/1c94589c7d9c1ff06a5220f775302e44f97a9956/src/epics/router.ts#L86-L96

AndrewRAM33: I believe that it necessary to update the BC in this epic.
It is true that if there are several filters, then the same BC will be updated the same number of times.
I fixed this by adding a change that would update only once each BC by which filter or sort with drilddown.


When `changeLocation` action is executed, we will no longer have an idea of ​​which BC filter was in the drilledown and which BC needs to be updated.
If try to associate a BC update with filter or sort in store, then problems will arise when drillDown with remove filter, or with filters for other BC that do not participate in drillDown with filtering.

Dergash:
After carefully examining the issue, the code and suggested fix I've come to the following conclusion:
1)  Too much time spent trying to understand what is there to fix, descriptions for both the issue and the pull requests are misleading and incomplete.
2) I believe the issue #374 concerns only one, very specific scenario: drilldown with explicit clearing of filters on the same view. This is due to existing branch with `bcForceUpdate` fires only for adding filters, but not for removing
3) Suggested changes ignore existing branch firing `bcForceUpdate` and instead append another `bcForceUpdate` call on top of it, which  definetly complicate the code and in effect also leadы to double fetching data when performing drilldown between the views sharing the BC.

I've adressed 2) in #451 and due to the 3) will close this PR.

PR #380, opened 2020-08-03
Add ability to exclude widget from being wrapped in <Card />

Guap6512:
Adds ability to exclude custom widgets from being wrapped in <Card /> component

### DIFF
--- a/src/components/Widget/Widget.tsx
+++ b/src/components/Widget/Widget.tsx
@@ -7,16 +7,16 @@ import {
     WidgetTypes,
     WidgetFormMeta,
     WidgetTableMeta,
-    CustomWidget,
     WidgetShowCondition,
-    WidgetTextMeta
+    WidgetTextMeta,
+    CustomWidgetDescriptor,
+    isCustomWidget,
 } from '../../interfaces/widget'
 import TableWidget from '../widgets/TableWidget/TableWidget'
 import TextWidget from '../widgets/TextWidget/TextWidget'
 import FormWidget from '../widgets/FormWidget/FormWidget'
 import styles from './Widget.less'
 import AssocListPopup from '../widgets/AssocListPopup/AssocListPopup'
-import {ObjectMap} from '../../interfaces/objectMap'
 import PickListPopup from '../widgets/PickListPopup/PickListPopup'
 import {BcMetaState} from '../../interfaces/bc'
 import {ViewState} from '../../interfaces/view'
@@ -32,7 +32,7 @@ interface WidgetOwnProps {
 interface WidgetProps extends WidgetOwnProps {
     loading?: boolean,
     parentCursor?: string,
-    customWidgets?: ObjectMap<CustomWidget>,
+    customWidgets?: Record<string, CustomWidgetDescriptor>,
     showWidget: boolean,
     rowMetaExists: boolean,
     dataExists: boolean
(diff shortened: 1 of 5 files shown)


src/components/Widget/Widget.tsx
```diff
@@ -44,7 +45,9 @@ export const Widget: FunctionComponent<WidgetProps> = (props) => {
     if (!props.showWidget) {
         return null
     }
-    if (props.meta.type === WidgetTypes.AssocListPopup || props.meta.type === WidgetTypes.PickListPopup) {
+    if (props.meta.type === WidgetTypes.AssocListPopup
+        || props.meta.type === WidgetTypes.PickListPopup
+        || props?.skipWrappingWidgets.includes(props.meta.type)) {
```
DRITE, line 50: I guess here should be something like this
```
|| props.skipWrappingWidgets?.includes(props.meta.type)) {
```

DRITE:
@Guap6512 there are  some problems now:
lint error -  https://github.com/tesler-platform/tesler-ui/pull/380/checks?check_run_id=944592395#step:4:8
UT error - https://github.com/tesler-platform/tesler-ui/pull/380/checks?check_run_id=944592314#step:4:11

Guap6512:
> @Guap6512 there are some problems now:
> lint error - https://github.com/tesler-platform/tesler-ui/pull/380/checks?check_run_id=944592395#step:4:8
> UT error - https://github.com/tesler-platform/tesler-ui/pull/380/checks?check_run_id=944592314#step:4:11

fixed

src/components/Widget/Widget.tsx
```diff
@@ -33,18 +33,24 @@ interface WidgetProps extends WidgetOwnProps {
...
 
+const SKIP_WRAPPING_WIDGETS: string[] = [WidgetTypes.AssocListPopup, WidgetTypes.PickListPopup]
+
 export const Widget: FunctionComponent<WidgetProps> = (props) => {
     if (!props.showWidget) {
         return null
     }
-    if (props.meta.type === WidgetTypes.AssocListPopup || props.meta.type === WidgetTypes.PickListPopup) {
+
+    const skipCardWrapping = SKIP_WRAPPING_WIDGETS.includes(props.meta.type) && props.skipWrappingWidgets?.includes(props.meta.type)
```
DRITE, line 51: Why conjunction has appeared here instead of disjunction?

Guap6512: fixed

Guap6512:
> 1. We need a broader approach to customization. If some widget types does not require a `<Card />` wrapper at all the next logical step is that different widget types will require a different `<Card />` implementations, and then we will stuck with over-complicated component props (skipWidget.
> 
> Try this instead:
> 
> ```tsx
> // Keep old behavior backward-compatible and extend it with optional config:
> /**
>  * TSDoc
>  */
> type CustomWidgetDescriptor = CustomWidget | {
>     /**
>      * TSDoc
>      */
>     component: CustomWidget,
> 
>     /**
>      * TSDoc
>      */
>     card?: CustomWidget
> }
> 
> export interface ViewProps {
>     customWidgets?: Record<string, CustomWidgetDescriptor>
> }
> ```
> 
> Then strict-check for `null`, truthy or `undefined` card value in `<Widget />` component to decide whether Card should be skipped, overridden or defaulted.
> 
> 1. There are two additional commits that does not belong to this PR

done

Dergash (approved):

Issue #376, opened 2020-08-03
`AutoSaveMiddleware` refactor


DRITE:
`AutoSaveMiddleware` should call saving for next cases:
- Dispatched action compatible to `needSaveAction` list;
- Dispatched action `selectTableCellInit` for another BC or another table row;
- Dispatched action `sendOperation` for another BC;
- Dispatched action `sendOperation` with type 'create'.

If result of one of those cases is true, then middleware should check being of any unsaved changes and only then call saving process.

PR #381, opened 2020-08-03
Refactored AutoSaveMiddleware (#376)

DRITE:
See #376 

It was decided not to follow #376's conditions.
By the way AutoSaveMiddleware has beed refactored with following conditions:
- Don't autosave if `operationType` is `save`;
- A saving performs for `sendOperation` actions called for another BC. But BCs having pending `_associate` are ignored;
- A saving performs for the action is `selectTableCellInit` called for another row or another widget;

### Details
+ The mechanism of autosaving has been extracted as separated utility to be exported from Tesler. So it can be used by custom middleware;
+ `needSaveAction` has been exported from tesler. So it can be used by custom middleware;
+ The mechanism of middleware customisation has been provided. It allows to set custom middlewares before, after or instead of core middlewares. Also core middlewares can be disabled.
For example:
```typescript
import {
CustomMiddlewarePriority,
 CustomMiddlewares,
 CoreMiddlewares
} from '@tesler-ui/core/interfaces/customMiddlewares'

// Describe an interface which extends `CoreMiddlewares`
interface MyMiddlewares extends Partial<CoreMiddlewares> {
    myFirstMiddleware: CustomMiddleware
    myLastMiddleware: CustomMiddleware
}

// Pass recently created interface to generic `CustomMiddlewares`
const myAwesomeMiddlewares: CustomMiddlewares<MyMiddlewares> = {

    // will be placed before tesler's middlewares
    myFirstMiddleware: { implementation: myFirstMiddlewareImpl, priority: CustomMiddlewarePriority.BEFORE },

    // will override tesler's autosave middelware
    autosave: myAutosaweMiddlewareOverriding,

    // will be placed after tesler's middlewares
    myLastMiddleware: { implementation: myLastMiddlewareImpl, priority: CustomMiddlewarePriority.AFTER },

    // teslers `requiredFields` middleware will be disabled
    requiredFields: null
}
```
Custom middlewares can be passed to tesler as `customMiddlewares` prop of tesler's `Provider`:
```typescript
import {Provider} from '@tesler-ui/core'
const App = <Provider
    customMiddlewares={myAwesomeMiddlewares}
    ... // other `Provider`'s props
>...</Provider>
```
### Addition
- Current  `Field` component behaviour shows `Tooltip` with error text above select field. In case select field opens dropdown above itself, tooltip will cover select options. So decided to enable ability to customise position of `Tooltip` by adding the prop `tooltipPlacement`. This prop takes values from antd's `TooltipPlacement` type.
- Component `AssocTable` isn't sending the whole data to `pendingDataChanges`


src/middlewares/autosaveMiddleware.ts
```diff
@@ -43,27 +84,64 @@ const saveFormMiddleware = ({ getState, dispatch }: MiddlewareAPI<Dispatch<AnyAc
...
              */
-
-            if (selectedCell
-                && (
-                    needToSaveTableChanges
-                    || (!isSelectTableCellInit && !isSendOperation)
-                    || (isSelectTableCellInit &&
-                        (selectedCell.widgetName !== action.payload.widgetName || selectedCell.rowId !== action.payload.rowId)
-                    )
+            const isNeedSaveCondition = isNotSaveAction &&
```
AndrewRAM33, line 89: The `sendOperation.associate` action from `actionGroup` to another BC causes autosave `OperationTypeCrud.save`. This should not happen because this operation calls the association widget popup. When performing an autosave in this case, the backend can respond with a `postInvoke` action to the autosave operation and the application will break.

AndrewRAM33 (changes requested):

src/middlewares/autosaveMiddleware.ts
DRITE: @AndrewRAM33 a mechanism of middleware overriding has been provided. Now you can implement required behaviour on client project.

src/utils/combineMiddlewares.ts
```diff
@@ -0,0 +1,53 @@
...
+            return
+        }
+        if (customMiddleware?.priority === CustomMiddlewarePriority.AFTER) {
+            customMiddlewaresAfter.push(customMiddleware.implementation)
+        }
+    })
+
+
+    // if there are present customMiddlewaresBefore then insert them before core middlewares
+    const resultMiddlewares: Middleware[] = customMiddlewaresBefore.length ? customMiddlewaresBefore : []
```
Dergash, line 34: It is almost identical to 
```ts
 const resultMiddlewares: Middleware[] = customMiddlewaresBefore
```
The only difference I can think of is that `[]` will give you are new array, but since you mutate original array in true-branch anyway It does not matter.

src/utils/combineMiddlewares.ts
```diff
@@ -0,0 +1,53 @@
...
+            resultMiddlewares.push(coreMiddlewares[coreMiddlewareName])
+            return
+        }
+        // Assigned custom implementation means it'll override the core implementation
+        if (customMiddleware) {
+            resultMiddlewares.push(customMiddleware as Middleware)
+            return
+        }
+    })
+    return customMiddlewaresAfter.length ? resultMiddlewares.concat(customMiddlewaresAfter) : resultMiddlewares
```
Dergash, line 52:
```ts
return [ ...resultMiddlewares, ...customMiddlewaresAfter ]
```

src/index.ts
```diff
@@ -182,3 +184,12 @@ export {
  * Epics implementations
  */
 export {sendOperationEpicImpl} from './epics/view'
+
+/**
+ * Autosave middleware utils
+ */
+export {
+    autosaveRoutine,
+    bcHasPendingAutosaveChanges,
```
Dergash, line 193: I do not think we should export `bcHasPendingAutosaveChanges` and `checkUnsavedChangesOfBc` as their are trivial to implement client-side.

src/interfaces/customMiddlewares.ts
```diff
@@ -0,0 +1,37 @@
...
+ */
+export interface CustomMiddleware {
+    /**
+     * Implementation of custom middleware
+     */
+    implementation: Middleware,
+    /**
+     * Priority of custom middleware
+     */
+    priority: CustomMiddlewarePriority
```
Dergash, line 22: String union probably will work better here:
```ts
priority: 'BEFORE' | 'AFTER'
```

Dergash (changes requested):

src/utils/autosave.ts
```diff
@@ -0,0 +1,97 @@
...
+
+/**
+ * Performs mechanism of autosave
+ *
+ * @param action
+ * @param state
+ * @param dispatch
+ * @param next
+ */
+export function autosaveRoutine(action: AnyAction, state: CoreStore, dispatch: Dispatch<AnyAction>, next: Dispatch) {
```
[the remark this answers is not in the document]
DRITE: fixed

src/utils/autosave.ts
```diff
@@ -0,0 +1,97 @@
...
+ * Checks presence of unsaved data for specified BC
+ *
+ * @param store
+ * @param bcName
+ */
+export function checkUnsavedChangesOfBc(store: CoreStore, bcName: string) {
+    let checkResult = false
+    const checkCursors = store.view.pendingDataChanges?.[bcName] && Object.keys(store.view.pendingDataChanges?.[bcName])
+    let i = 0
+    for(; i < checkCursors?.length; i++) {
```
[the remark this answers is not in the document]
DRITE: Aye. It was silly. Fixed.

src/utils/combineMiddlewares.ts
DRITE: @Dergash I have not got what 'original array' you are talking about.
This is an initialisation of `resultMiddlewares`. So it is either `customMiddlewaresBefore` or a new array.
But anyway, changed to:
```typescript
const resultMiddlewares: Middleware[] = customMiddlewaresBefore
```

src/utils/combineMiddlewares.ts
DRITE: Fixed

src/index.ts
DRITE: Ok

src/interfaces/customMiddlewares.ts
DRITE: Fixed

src/interfaces/customMiddlewares.ts
```diff
@@ -0,0 +1,37 @@
...
+}
+
+/**
+ * Descriptor of custom middleware not presented in core middlewares
+ */
+export type NewMiddlewareDescriptor = {[middlewareName: string]: CustomMiddleware}
+/**
+ * Descriptor of core middlewares
+ */
+export type CoreMiddlewareDescriptor = {[middlewareName: string]: Middleware}
```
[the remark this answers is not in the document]
DRITE: Thank you
Fixed

src/interfaces/customMiddlewares.ts
```diff
@@ -0,0 +1,37 @@
...
+
+/**
+ * Type of custom middlewares
+ */
+export type CustomMiddlewares = Partial<Record<keyof typeof middlewares, Middleware | null>> | NewMiddlewareDescriptor
```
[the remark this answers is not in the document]
DRITE: Thank you for extended comment.
I've refactored types new way:
```typescript
/**
 * Type of core middlewares
 */
export type CoreMiddlewares = typeof middlewares

/**
 * Custom middleware interface
 */
export interface CustomMiddleware {
    /**
     * Implementation of custom middleware
     */
    implementation: Middleware,
    /**
     * Priority of custom middleware
     */
    priority: 'BEFORE' | 'AFTER'
}

/**
 *  List the names of all core middlewares
 */
export type CoreMiddlewareType = keyof CoreMiddlewares

/**
 * Descriptor of custom middleware not presented in core middlewares
 */
export type NewMiddlewareDescriptor<T = {}> = Record<Exclude<keyof T, keyof CoreMiddlewares>, CustomMiddleware>

/**
 * Form a dictionary of override descriptors for those middleware
 */
export type CoreMiddlewareOverrideDescriptors = Record<CoreMiddlewareType, Middleware | null>
/**
 * Type of custom middlewares
 */
export type CustomMiddlewares<T = {}> = Partial<CoreMiddlewareOverrideDescriptors> | NewMiddlewareDescriptor<T>
```
It provides usage as described in PR's description.

Dergash (approved):

Issue #363, opened 2020-07-29
There is need ability to set maximum length to input field


DRITE:

DRITE:
This feature is added in [1.19.0](https://github.com/tesler-platform/tesler-ui/milestone/49)

PR #385, opened 2020-08-04
Added 'maxInput' param to WidgetFieldBase (#363)

DRITE:
see #363 

Added availability to set maximum number of characters to fields of type `input` and `text`.

Upgraded antd to 3.26.18 in aims of fixing https://github.com/ant-design/ant-design/issues/21697

### DIFF
--- a/src/components/ui/TextArea/TextArea.tsx
+++ b/src/components/ui/TextArea/TextArea.tsx
@@ -25,7 +25,7 @@ export interface TextAreaProps extends AdditionalAntdTextAreaProps {
     onDrillDown?: () => void
 }
 
-const TextArea: React.FunctionComponent<TextAreaProps> = (props) => {
+export const TextArea: React.FunctionComponent<TextAreaProps> = (props) => {
     if (props.readOnly) {
         return <ReadOnlyField
             className={props.className}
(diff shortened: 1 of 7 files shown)


src/components/ui/TextArea/TextArea.tsx
```diff
@@ -8,11 +8,16 @@ import InputDefaultClass from 'antd/lib/input/TextArea'
...
 import ReadOnlyField from '../ReadOnlyField/ReadOnlyField'
 import {TextAreaProps as AntdTextAreaProps} from 'antd/lib/input/TextArea'
+import {WidgetField} from 'interfaces/widget'
 
 type AdditionalAntdTextAreaProps = Partial<Omit<AntdTextAreaProps, 'onChange'>>
 export interface TextAreaProps extends AdditionalAntdTextAreaProps {
+    cursor?:string,
+    widgetName?: string,
+    metaError?: string,
+    meta?: WidgetField, // TODO 2.0.0 must be required?
```
Dergash, line 18: It unnecessary binds a regular UI component to a widget context and narrows its usage, i.e. you'll no longer be able to use this component in other parts of the application.  
UI component shouldn't care about widgets, cursors and meta.

src/components/ui/TextArea/TextArea.tsx
```diff
@@ -8,11 +8,16 @@ import InputDefaultClass from 'antd/lib/input/TextArea'
 import styles from './TextArea.less'
 import ReadOnlyField from '../ReadOnlyField/ReadOnlyField'
 import {TextAreaProps as AntdTextAreaProps} from 'antd/lib/input/TextArea'
+import {WidgetField} from 'interfaces/widget'
```
Dergash, line 11: Should be relative

src/interfaces/widget.ts
```diff
@@ -31,6 +31,10 @@ export const TableLikeWidgetTypes: Array<WidgetTypes | string> = [
     WidgetTypes.PickListPopup
 ]
 
+/**
+ * @param maxInput Maximum number of characters
+ * @param hidden Whether the field is hidden
```
Dergash, line 36: TSDoc for interface properties should be placed separately for each corresponding property, i.e. `@param` is not valid for interfaces and will not be parsed correctly.

See reference:
https://github.com/microsoft/tsdoc/blob/746a8fb8fe89790107325c79e9077cdc8c83546a/playground/src/CodeEditor.tsx#L5

DRITE: I did as you are https://github.com/tesler-platform/tesler-ui/commit/2006015d86bdc0b76af3f9a1f492c877eb30a19e#diff-212d66cacab7e2fade02457e67695c21R132
I'll fix it

Dergash (approved):

Issue #387, opened 2020-08-04
Custom action autosave


AndrewRAM33:
Widget custom action as default save action. Used in `autosaveMiddleware.ts`.

PR #388, opened 2020-08-04
Custom action autosave

AndrewRAM33:
Custom action autosave (#387)

**Feature:** Autosave pending data by custon widget save action.

New ability to autosave by custom action on form widgets when have only custom actions. Required when user has input form and initiated drilldown.

New widget meta param `defaultSave` is name default no crud save action
_**Example:**_
``` ts
  "name": "Widget Name",
  "title": "Widget Title",
  "type": "Form",
  "bc": "bcName",
  ...
  "options": {
    "actionGroups": {
      "include": [
        "custom-save",
        "custom-cancel"
      ],
      "exclude": [
        "custom-edit",
      ],
      "defaultSave": "custom-save"
    },
   ...
  },
  }
```
When user input part or full Form and initial any `drillDown`, action `"defaultSave": "custom-save"` will be executed. But if custom save action `"defaultSave": "custom-save"` have `postAction` `drillDown`, then `postAction` not be perfomed, execute only user `drillDown`.

### DIFF
--- a/src/interfaces/operation.ts
+++ b/src/interfaces/operation.ts
@@ -309,13 +309,14 @@ export interface AssociatedItem extends DataItem {
 }
 
 /**
- * Дескриптор включения операции на виджете:
- * - либо строка (если надо просто включить/исключить операцию или группы)
- * - либо объект, если это группа в которой нужно выборочно включить или исключить операцию
+ * Descriptor enabling operation on widget:
+ * - string (if you just need to include / exclude operation or groups)
+ * - object, if this is group in which you want to selectively include or exclude the operation
  * 
- * @param type Тип операции; строка, уникально идентифицирующая операцию на виджете
- * @param include Список включаемых операций
- * @param exclude Список исключаемых операций
+ * @param type Type of transaction; a string that uniquely identifies the operation on the widget
+ * @param include List of included operations or groups operations
+ * @param exclude List of excluded operations or groups operations
+ * @param defaultSave default no crud save action
  */
 export type OperationInclusionDescriptor = string | {
     type: OperationType,
--- a/src/middlewares/autosaveMiddleware.ts
+++ b/src/middlewares/autosaveMiddleware.ts
@@ -21,6 +21,32 @@ const saveFormMiddleware = ({ getState, dispatch }: MiddlewareAPI<Dispatch<AnyAc
             const needToSaveTableChanges = isSendOperation && isAnotherBc
             const selectedCell = state.view.selectedCell
             const isSelectTableCellInit = action.type === types.selectTableCellInit
+
+
+            /**
+             * Default save operation as custom action
+             *
+             * If widget have only custom actions, `defaultSave` option mean witch action
+             * must be executed as save record.
+             * Current changeLocation action as onSuccessAction
+             */
+            const defaultSaveWidget = state.view.widgets?.find(item => item?.options?.actionGroups?.defaultSave)
+            const defaultCursor = state.screen.bo.bc?.[defaultSaveWidget?.bcName]?.cursor
+            const pendingData = state.view?.pendingDataChanges?.[defaultSaveWidget?.bcName]?.[defaultCursor]
+            if (defaultSaveWidget && action.type === types.changeLocation && pendingData) {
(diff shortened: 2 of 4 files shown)


src/interfaces/operation.ts
```diff
@@ -309,18 +309,20 @@ export interface AssociatedItem extends DataItem {
...
+ * @param include List of included operations or groups operations
+ * @param exclude List of excluded operations or groups operations
+ * @param defaultSave default no crud save action
  */
 export type OperationInclusionDescriptor = string | {
     type: OperationType,
     include?: OperationInclusionDescriptor[],
-    exclude?: OperationType[]
+    exclude?: OperationType[],
+    defaultSave?: string
```
Dergash, line 325: I don't think it is a correct place to configure operations:
* Type name is misleading (`OperationInclusionDescriptor` gives no hint that it also handles autosaves)
* `OperationInclusionDescriptor` is used for optional `include` config:
https://github.com/tesler-platform/tesler-ui/blob/1c94589c7d9c1ff06a5220f775302e44f97a9956/src/interfaces/widget.ts#L303-L311
It means that if you want to specify autosave action you also should configure widget operations via `include` property and autosave can't be used if your operations configured with `exclude` or does not care about inclusion/exclusion logic at all

`WidgetOperations` should be used for specifying default save action instead.

src/middlewares/autosaveMiddleware.ts
```diff
@@ -21,6 +21,35 @@ const saveFormMiddleware = ({ getState, dispatch }: MiddlewareAPI<Dispatch<AnyAc
...
+             * If widget have only custom actions, `defaultSave` option mean witch action
+             * must be executed as save record.
+             * Work only user initial `changeLocation` and don't work at postAction `drillDown`
+             * @param ignorePostAction Indicate that postAction not needed (postAction.drillDown as example)
+             */
+            const defaultSaveWidget = state.view.widgets.find(item => item?.options?.actionGroups?.defaultSave)
+            const defaultCursor = state.screen.bo.bc?.[defaultSaveWidget?.bcName]?.cursor
+            const pendingData = state.view?.pendingDataChanges?.[defaultSaveWidget?.bcName]?.[defaultCursor]
+            if (defaultSaveWidget && action.type === types.changeLocation && !action.ignoreAutosave && pendingData) {
+                action.ignoreAutosave = true
```
Dergash, line 38: I think `ignoreAutosave` essentially means "all save operations should not fire autosave middleware".  
Try `actionRole: 'save'` for custom autosave actions and `matchOperationRole` to check for early exit from middleware:
https://github.com/tesler-platform/tesler-ui/blob/1c94589c7d9c1ff06a5220f775302e44f97a9956/src/utils/operations.ts#L43-L52

src/middlewares/autosaveMiddleware.ts
```diff
@@ -21,6 +21,35 @@ const saveFormMiddleware = ({ getState, dispatch }: MiddlewareAPI<Dispatch<AnyAc
...
+                return next($do.sendOperation({
+                    bcName: defaultSaveWidget.bcName,
+                    operationType: defaultSaveWidget.options.actionGroups.defaultSave,
+                    widgetName: defaultSaveWidget.name,
+                    onSuccessAction: action,
+                    ignorePostAction: true
```
Dergash, line 44: I believe `ignorePostAction` will not be necessary if you use `actionRole: 'save'` because `sendOperation` epic that handles post invokes won't fire for cruds

src/interfaces/operation.ts
AndrewRAM33: removed

src/middlewares/autosaveMiddleware.ts
AndrewRAM33: removed

src/middlewares/autosaveMiddleware.ts
AndrewRAM33: resolved with `defaultSave` widget meta option and without `ignoreAutosave`, `ignorePostAction` flags.

Dergash (approved):

Issue #389, opened 2020-08-06
Show error on ReadOnly field in TableWidget


sergromanov4:
An error may occur when validating field changes in `TableWidget`. And for the `ReadOnly` field this error is not displayed in any way.

Although when an error occurs while in `editmode` the border of the field turns red and when the cursor hovers over it a `tooltip` appears with a description of the error.

PR #390, opened 2020-08-06
ReadOnly field show error on TableWidget

sergromanov4:
see (https://github.com/tesler-platform/tesler-ui/issues/389)

### DIFF
--- a/src/components/Field/Field.tsx
+++ b/src/components/Field/Field.tsx
@@ -233,6 +233,7 @@ export const Field: FunctionComponent<FieldProps> = (props) => {
                 {...commonProps}
                 defaultValue={value as any}
                 onChange={handleChange}
+                className={cn({[readOnlyFieldStyles.error]: props.metaError})}
             />
             break
         case FieldType.multifield:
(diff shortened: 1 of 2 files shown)


src/components/ui/TextArea/TextArea.tsx
```diff
@@ -22,12 +22,14 @@ export interface TextAreaProps extends AdditionalAntdTextAreaProps {
     maxRows?: number,
     className?: string,
     backgroundColor?: string,
-    onDrillDown?: () => void
+    onDrillDown?: () => void,
+    metaError?: string
```
Dergash, line 26: As in the previous scenario `<TextArea />` should rely on the parent `<Field />` component to handle errors and pass an appropriate `className`.  
One example of reasoning behind this is that there are many others components aside of `<TextArea />` which reuse `<ReadOnlyField />`:
* `<RadioButton />`
* `<NumberInput />`
* `<Dictionary />`
* `<DatePickerField />`
* `<PickListField />`
* `<InlinePickList />`

Dergash (approved):

Issue #396, opened 2020-08-10
<Dictionary/>: multiple mode is needed


DRITE:
Decided to refactor implementation of  `Dictionary` component.

Example of ui usage:
```case FieldType.dictionary:
    resultField = <Dictionary
        {...commonProps}
        multiple={props.widgetFieldMeta.multiple}
        onChange={handleChange}
    />
```
Example of meta:
```
    {
      "label": "Language",
      "key": "languageList",
      "type": "dictionary",
      "multiple": true
    }
```

DRITE:
This feature in [1.19.0](https://github.com/tesler-platform/tesler-ui/milestone/49)

PR #397, opened 2020-08-10
<Dictionary/>: added `multiple` mode for several items selecting

DRITE:
See #396

### DIFF
--- a/src/components/Field/Field.tsx
+++ b/src/components/Field/Field.tsx
@@ -226,6 +226,7 @@ export const Field: FunctionComponent<FieldProps> = (props) => {
                 values={props.rowFieldMeta ? props.rowFieldMeta.values : []}
                 fieldName={props.widgetFieldMeta.key}
                 onChange={handleChange}
+                multiple={props.widgetFieldMeta.multiple}
             />
             break
         case FieldType.text:
--- a/src/components/ui/Dictionary/Dictionary.test.tsx
+++ b/src/components/ui/Dictionary/Dictionary.test.tsx
@@ -1,0 +1,163 @@
+import {shallow} from 'enzyme'
+import Dictionary, {DictionaryProps, getIconByParams} from './Dictionary'
+import * as React from 'react'
+import {MultivalueSingleValue} from 'interfaces/data'
+
+const defValues = [
+    {value: 'German'},
+    {value: 'Danish'},
+    {value: 'Low German'},
+    {value: 'Sorbian'},
+    {value: 'Romany'},
+    {value: 'Frisian'},
+    {value: 'English'},
+    {value: 'Scots'},
+    {value: 'Welsh'},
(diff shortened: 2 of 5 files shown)


src/components/ui/MultipleSelect/MultipleSelect.tsx
```diff
@@ -0,0 +1,83 @@
...
+                )
+            })
+        },
+        [values]
+    )
+
+    const handleOnChange = React.useCallback(
+        (v: string[]) => {
+            const result: MultivalueSingleValue[] = []
+            v.map(item => result.push({id: item, value: item}))
```
Dergash, line 61: It's either
```tsx
const result = v.map(item => ({ id: item, value: item }))
```
or
```tsx
const result: MultivalueSingleValue[] = []
v.forEach(item => {
    result.push({ id: item, value: item })
})
```

There is no point in mapping if the result is not assigned to something: essentially you've mapped `v` to a new array `[1, 2, 3, ..., v.length - 1]` and then dropped it.

src/components/ui/MultipleSelect/MultipleSelect.tsx
```diff
@@ -0,0 +1,83 @@
...
+    className: string,
+    metaError: string,
+    disabled: boolean,
+    placeholder: string,
+    readOnly: boolean,
+    backgroundColor: string,
+    onDrillDown: () => void,
+}
+
+const MultipleSelect: React.FunctionComponent<MultipleSelectProps> = (props) => {
```
Dergash, line 26: I can't understand the difference with our regular `<Dictionary />` component, implementation looks almost identical aside of `multiple` option passed to the underlying antd `<Select />` and some missing features.

Wouldn't it be easier to make something like this:
```tsx
case FieldType.dictionary:
    resultField = <Dictionary
        {...commonProps}
        multiple={props.widgetFieldMeta.multiple}
        onChange={handleChange}
    />
```
We do support custom options for fields description in widget meta.

DRITE: @Dergash 
I thought about that. Here some reasons why decided to implement an another component:
- New component has simple implementation.
- As `Dictionary` component can get single value it allows to `drilldown` its value. This behavior is not suitable for `Select` in `multiple` mode.
- Extending `Dictionary`'s implementation leads to increasing complexity of  its testing and overriding, I think.

DRITE: @Dergash 
I've got it. I'll replace it by refactored `Dictionary`.
But I think its usage will look like:
```
case FieldType.multipleSelect:
    resultField = <Dictionary
        {...commonProps}
        multiple
        onChange={handleChange}
    />
```
Ok?

Dergash (approved):

Issue #434, opened 2020-09-09
Custom autosave bugs


AndrewRAM33:
Custom autosave middleware have some bugs:
1. Execute when `pendingData` is a empty object.
2. Execute when receive `drillDown` `postAction` and exist `pendingData`.

Dergash:
Solved in #476

PR #435, opened 2020-09-09
Custom autosave bugfix

AndrewRAM33:
See #434 

Bugfix custom autosave:
1. Check `pendingData` as not empty object with `Object.keys(pendingData).length` function
2. Execute `bcCancelPendingChanges` action before `drillDown` `postAction`.  This will clear `pendingData` before `changeLocation`. So custom autosave not perform, if exist.

### DIFF
--- a/src/epics/screen.ts
+++ b/src/epics/screen.ts
@@ -18,11 +18,17 @@ const processPostInvoke: Epic = (action$, store) => action$.ofType(types.process
     const state = store.getState()
     switch (action.payload.postInvoke.type) {
         case OperationPostInvokeType.drillDown:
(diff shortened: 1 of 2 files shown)


Dergash (changes requested):
Autosave middleware exists precisely to caught unsaved pending changes that will be lost otherwise; discarding pending changes on drilldown leads to arbitrary and illogical behavior (why we drop pending changes on drilldown but not on location change?).
It should not be default behavior, client side application should provide its own middleware which will be catching postInvoke type and firing cancel pending dara

AndrewRAM33:
closed in favor #476

Issue #437, opened 2020-09-11
Short text fields (hidden tail)


AndrewRAM33:
Need to be able to render long text fields incompletely.
Also make it possible to display the full field and return to its original state.

**_Example:_**
_Original_ 
``` ts
Very long text field does not fit in the cell
```

_Expect_ 
``` ts
Very long text field...
```

_Expect expand_ 
``` ts
Very long text field does not fit in the cell'
(hide again button)
```

PR #438, opened 2020-09-11
Draft: Short text fields (hidden tail) 

AndrewRAM33:
See #437 

Create new component `HiddenString`. This component allow replace any string value at short string with expand to full string button as '...'. After expand you have ability to hide again with button 'Hide'. Both buttons are implemented as plain clickable text with pointer cursor.

This replace is active when field have prop `showLength?: number`. 

Modification is only available for text like field types:
- text
- hint
- input (readOnly)

**_Example:_**
_Original_ 
``` ts
Very long text field does not fit in the cell
```

_Expect_ 
``` ts
Very long text field...
```

_Expect expand_ 
``` ts
Very long text field does not fit in the cell'
(hide again button)
```


AndrewRAM33:
New implementation in client APP

Issue #441, opened 2020-09-12
`pendingValidationFails` should be divided by BC names and cursors


DRITE:
There is incorrect behaviour of validation in next cases:
- Two (or more) widgets includes fields with same keys (all these fields get validation error)
- Any table widget (a whole column get validation error)

PR #442, opened 2020-09-12
`pendingValidationFails` are divided by BC names and cursors (#441)

DRITE:
See #441 

A target form of the interface of `pendingValidationFails` is looks like:
```
    pendingValidationFails?: PendingValidationFails
```
where `PendingValidationFails` is:
```
interface PendingValidationFails {
    [bcName: string]: {
        [cursor: string]: Record<string, string>
    }
}

```

### Backward compatibility
I provide temporary solution for backward compatibility until `2.0.0` release will be released.
Temporary interface for `pendingValidationFails` looks like this:
```
pendingValidationFails?: Record<string, string> | PendingValidationFails,
```  
I've added `pendingValidationFailsFormat` as flag of `target` or `old` formats of `pendingValidationFails`.
Currently, if you'd like to use target version of `pendingValidationFails`, you can simply set
```pendingValidationFailsFormat: PendingValidationFailsFormat.target``` in initial state of overriding of `view` reducer.

### DIFF
--- a/src/components/Field/Field.tsx
+++ b/src/components/Field/Field.tsx
@@ -5,7 +5,7 @@ import {Form, Icon, Input, Tooltip} from 'antd'
 import {$do} from '../../actions/actions'
 import {Store} from '../../interfaces/store'
 import {DataItem, DataValue, MultivalueSingleValue, PendingDataItem} from '../../interfaces/data'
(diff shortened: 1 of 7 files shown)


DRITE:
I've leaved two commits. The second commit gives more simple way to look for code which must be deleted in `2.0.0` release.

Issue #443, opened 2020-09-14
HierarchyTraverse diselection bug


AndrewRAM33:
Bug was found in the operation of diselect `FullHierarchy` item with `hierarchyTraverse` option.

See #5
- **hierarchyTraverse** - Select/Deselect all ancestors up to root when record is selected/deselected

Diselect item action match ancestor selection at last unselected item in `Data` and `Delta`. 
The error is that for different items on same level the selection uses the same date, which does not change when select/diselect another items on this level.

The parent will not diselected if another item is selected in the child data item, then deselected both, and first from the one that has already been selected.

_**Example:**_
```
Default
P|      selected 
 |-C1   selected
 |-C2   unselected
```
1. open `FullHierarcy` with this configuration and `hierarchyTraverse = true`
2. select/unselect C1 was change P
3. select C2
4. unselect C1
5. unselect C2
Parent was selected and don't have selected child. Bug.

PR #444, opened 2020-09-15
hierarchyTraverse diselect bugfix. MVG field hierarchy behavior.

AndrewRAM33:
See #443 

`hierarchyTraverse` diselect bugfix. 

Additional filter for `wasLastInData` has been added to `currentLevelData` check, which excludes from `currentLevelData` array those that have been diselected.

MVG field tags will inherit behavior delete items from the hierarchy widget with which it is associated. Unavailable close option before AssocPopup is ready.

**_bugfix:_**
- Application crach when `AssocPopup` widget is not Hierarchy without `meta.options`
- Fix match popup pendingChanges for not Hierarchy `AssocPopup`

### DIFF
--- a/src/components/widgets/AssocListPopup/AssocListPopup.tsx
+++ b/src/components/widgets/AssocListPopup/AssocListPopup.tsx
@@ -27,7 +27,8 @@ export interface IAssocListRecord {
 export interface IAssocListActions {
     onSave: (bcNames: string[]) => void,
     onFilter: (bcName: string, filter: BcFilter) => void,
-    onDeleteTag: (bcName: string, dataItem: DataItem) => void,
+    onDeleteTag: (bcName: string, depth: number, widgetName: string,
+                  dataItem: AssociatedItem, assocValueKey: string) => void,
     onDeleteAssociations: (bcName: string, parentId: string, depth: number, assocValueKey: string, selected: boolean) => void,
     onRemoveFilter?: (bcName: string, filter: BcFilter) => void,
     onCancel: () => void,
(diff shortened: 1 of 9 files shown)


src/components/widgets/AssocListPopup/AssocListPopup.tsx
```diff
@@ -27,7 +27,8 @@ export interface IAssocListRecord {
 export interface IAssocListActions {
     onSave: (bcNames: string[]) => void,
     onFilter: (bcName: string, filter: BcFilter) => void,
-    onDeleteTag: (bcName: string, dataItem: DataItem) => void,
+    onDeleteTag: (bcName: string, depth: number, widgetName: string,
```
Dergash, line 30: 1) Adding new arguments brokes the contract for this interface
2) You don't really need two arguments that you've add: `depth` is just an alias for `dataItem.level` and `dataItem` is already an argument, and you can get `widgetName` in epic from the `store.view.popupData`

src/epics/data.ts
```diff
@@ -965,22 +970,69 @@ const removeMultivalueTag: Epic = (action$, store) => action$.ofType(types.remov
...
+        const removedItem = action.payload.dataItem.find(item => item.id === parentItem.id)
+        const currentLevelData = allData?.filter(
+            (item) => item.level === removeDataItem?.level && (item.level === 1 || (item.parentId === removeDataItem?.parentId))
+        )
+        const currentLevelDelta = delta.filter(
+            (item) => item.level === removeDataItem?.level && (item.level === 1 || (item.parentId === removeDataItem?.parentId))
+        )
+        const wasLastInDelta = !currentLevelDelta
+            .filter(item => item.id !== removeDataItem.id)
+            .filter(item => item.level === removeDataItem.level).length
```
Dergash, line 984: Can be combined with previous `filter`

Dergash (changes requested):

src/epics/data.ts
```diff
@@ -970,22 +975,69 @@ export const removeMultivalueTag: Epic = (action$, store) => action$.ofType(type
...
+    }
+
+    if (widget.options?.hierarchyGroupDeselection) {
+        // delta include only associated items and contain associated dataItems
+        const currentData = delta.length > 0 ? [] : allData.filter(item => item._associate === true)
+        const currentDataItems = [...currentData, ...delta]
+        const childDataItems = currentDataItems.filter(item => item.parentId === removeDataItem.id)
+        const deleteItems: DataItem[] = []
+        while(childDataItems.length > 0){
+            const tempItem = childDataItems.shift()
```
Dergash, line 1014: This approach does not match `hierarchyTraverse` option: it'll fetch associated descendants of associated nodes, but fill not fetch associated descendants of unassociated nodes.  

I think without `hierarchyTraverse` option it should only fetch direct children; otherwise it should fetch all associated descdendants, without skipping intermidiate unassociated nodes

Dergash (changes requested):

Issue #445, opened 2020-09-15
Broken `coreInitialState` initialization


DRITE:
There is a bug while `coreInitialState` initialization.
When we combine core initial state with custom initial state core's state doesn't take part in combination.
```
const coreInitialState = coreReducers[reducerName]
const reducerInitialState = {
    ...(coreInitialState || {} as ClientState),
    ...customReducers[reducerName].initialState,
    }
```
```coreReducers[reducerName]``` is not initial state, it's reducer itself.
So `reducerInitialState` is just a copy of `customReducers[reducerName].initialState`

PR #446, opened 2020-09-15
Bug of combination core and custom initial states is fixed (#445)

DRITE:
see #445

### DIFF
--- a/src/Provider.tsx
+++ b/src/Provider.tsx
@@ -105,7 +105,7 @@ export function configureStore<ClientState, ClientActions extends Action<any>>(
     // TODO: Extract this to an utility
     const reducers = { ...coreReducers } as CombinedReducersMapObject<CoreStore & ClientState, CombinedActions>
     Object.keys(customReducers).forEach((reducerName: Extract<keyof ClientState, string>) => {
-        const coreInitialState = coreReducers[reducerName]
+        const coreInitialState = coreReducers[reducerName]?.(undefined, {type: ' UNKNOWN ACTION '})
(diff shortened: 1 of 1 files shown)


src/Provider.tsx
```diff
@@ -105,7 +105,7 @@ export function configureStore<ClientState, ClientActions extends Action<any>>(
     // TODO: Extract this to an utility
     const reducers = { ...coreReducers } as CombinedReducersMapObject<CoreStore & ClientState, CombinedActions>
     Object.keys(customReducers).forEach((reducerName: Extract<keyof ClientState, string>) => {
-        const coreInitialState = coreReducers[reducerName]
+        const coreInitialState = coreReducers[reducerName](undefined, {type: ' UNKNOWN ACTION '})
```
Dergash, line 108: Null coalescing should be applied or we'll crash the app when we process a reducer slice which is not part of our core (store.getState().customSlice). 

Good catch otherwise. I'm unsure how it should've been written to avoid this kind of bug:  functions are object type hence allowed to be spread, and it seems that customReducers' initial state pass typings due to extending nature of the client slice over the core slice. I'd say overall typings for the provider should be improved.

Dergash (changes requested):

src/Provider.tsx
DRITE: Changed to
```const coreInitialState = coreReducers[reducerName]?.(undefined, {type: ' UNKNOWN ACTION '})```

Dergash (approved):

PR #451, opened 2020-09-22
More clear epic for handling drilldown filters, do not put empty filters in the store (#374)

Dergash:
Rewrite for handling inner drilldowns with filters:
1) Existing filters are cleared only when drilldown specifies its own filters or explicitly says that filters should be cleared via empty string filter expression. Otherwise filters are preserved, the same way as they are preserved when navigating between views via menu.
2) Drilldown to the same view with explicit clearing of filters now should correctly fire `bcForceUpdate` (adressess #374)
3) Empty filters now completely removed from the store instead of hanging with empty objects leading to various checks throughout the code thinking that filters are set for the BC

### DIFF
--- a/src/epics/router.ts
+++ b/src/epics/router.ts
@@ -26,7 +26,7 @@ const changeLocation: Epic = (action$, store) => action$.ofType(types.changeLoca
 .mergeMap(action => {
     const state = store.getState()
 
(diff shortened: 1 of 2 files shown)


AndrewRAM33 (approved):
Good decision. Overrides fix #375

PR #457, opened 2020-09-30
Cancel proccessing 'fetchBcData' request after logout

Guap6512:
See #456


Dergash:
@Guap6512  I see no harm in merging this but underlying problem is a bit deeper: it only handles data requests, but you also have row meta requests which still might fire after logout, and action requests, and many others.  
Additionally there many other places aside of logout where this should be implemented, so to solve #456 we need a concept of request cancellation (both for epics and ajax calls).

Issue #459, opened 2020-10-01
AssocListPopup and PickListPopup visibility


Guap6512:
AssocListPopup and PickListPopup show prop is treated differently, AssocListPopup is rendered only when props.show is true and PickListPopup is rendered always, but shown only when props.show is true. This behaviour needs to be unified

PR #460, opened 2020-10-01
Unify PickListPopup and AssocListPopup showed prop

Guap6512:
#459

### DIFF
--- a/src/components/widgets/AssocListPopup/AssocListPopup.tsx
+++ b/src/components/widgets/AssocListPopup/AssocListPopup.tsx
@@ -44,7 +44,11 @@ export interface IAssocListOwnProps extends Omit<PopupProps, 'bcName' | 'childre
 }
 
 export interface IAssocListProps extends IAssocListOwnProps {
-    showed: boolean,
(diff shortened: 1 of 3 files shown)


DRITE (approved):

src/components/widgets/AssocListPopup/AssocListPopup.tsx
```diff
@@ -44,7 +44,6 @@ export interface IAssocListOwnProps extends Omit<PopupProps, 'bcName' | 'childre
 }
 
 export interface IAssocListProps extends IAssocListOwnProps {
-    showed: boolean,
```
Dergash, line 47: This will break all client apps that use this component: any changes to public API only allowed in major release, so instead of removal it should be marked as optional and deprecated:
```ts
/**
 * @deprecated TODO: Remove in 2.0.0, now handled by Widget.tsx
*/
showed?: boolean
```

Dergash (changes requested):

Dergash (approved):

Issue #507, opened 2020-11-09
Need to update `OperationGroup` interface


DRITE:
As you van see in https://github.com/tesler-platform/tesler/commit/6dd4648d41440e2dbde86dd963ad05be2fa90dd2
It's become possible to set icon to operation group

So need to add `icon` and `showOnlyIcon` to interface `OperationGroup`

PR #508, opened 2020-11-09
`icon` and `showOnlyIcon` are added to interface OperationGroup (#507)

DRITE:
see #507

### Addition
`calleeWidgetName` is added to interface PopupData

### DIFF
--- a/src/interfaces/operation.ts
+++ b/src/interfaces/operation.ts
@@ -98,46 +98,69 @@ export interface Operation {
 }
 
 /**
+ * Group of actions.
+ *
+ * It shows name of a group, drop down list of actions
+ * and some actions which are shown in case list is covered.
  * Группа действий, показывает название группы и раскрываемые список ее действий,
  * а также несколько действий рядом с группой, которые видны не раскрывая список.
- *
- * @param text - отображаемое название группы действий
- * @param actions - список действий в группе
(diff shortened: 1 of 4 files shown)


src/interfaces/operation.ts
```diff
@@ -98,74 +98,99 @@ export interface Operation {
...
  * Группа действий, показывает название группы и раскрываемые список ее действий,
  * а также несколько действий рядом с группой, которые видны не раскрывая список.
- *
- * @param text - отображаемое название группы действий
- * @param actions - список действий в группе
- * @param maxGroupVisualButtonsCount - сколько действий будут видны, не раскрывая список
  */
 export interface OperationGroup {
+    /**
+     * ???
```
Dergash, line 110: It's meant as unique identifier for the operation group; it is used by [widget.options.actionGroups](https://github.com/tesler-platform/tesler-ui/blob/b3f517597e6d52ecffa49524f7632700244edabd/src/interfaces/widget.ts#L237) and [useWidgetOperation](https://github.com/tesler-platform/tesler-ui/blob/master/src/hooks/useWidgetOperations.ts) to decide which group of operations to include or exclude, as operations groups configured in business component service and widgets sharing the same BC might need a different set of available actions/action groups

Dergash (approved):

PR #529, opened 2020-12-08
ErrorPopup should support multiline text

DRITE:


src/components/ui/ErrorPopup/ErrorPopup.less
```diff
@@ -2,6 +2,10 @@
     display: flex;
     flex-direction: column;
 
+    :global(.ant-form-item-children) {
```
Dergash, line 5: Can we assign this logic to a separate class?

DRITE: @Dergash Do you meat something like this?
https://github.com/tesler-platform/tesler-ui/pull/529/commits/2754db011c5efef07137fe9c487d69c188999e58#diff-7845ebc39b1c85dac4e17f911d6971d153be683b6b9dca7014efb6e4ebd297feR62

Dergash (approved):

PR #535, opened 2020-12-16
Ability to override drilldown link by adding drillDownKey in row meta

Lourence94:
Url can be taken from field data by given drillDownKey, otherwise stays default drilldown behaviour

### DIFF
--- a/src/hooks/useDrillDownUrl.ts
+++ b/src/hooks/useDrillDownUrl.ts
@@ -1,0 +1,29 @@
+import {useSelector} from 'react-redux'
+import {Store} from '../interfaces/store'
+import {WidgetFieldBase} from '../interfaces/widget'
+import {buildBcUrl} from '../utils/strings'
+
+/**
+ * @param widgetName - widget name passed to field
+ * @param cursor - field ID
+ * @param fieldMeta - widget field meta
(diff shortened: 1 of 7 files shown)


src/hooks/useDrillDownUrl.ts
```diff
@@ -0,0 +1,28 @@
+import {useSelector} from 'react-redux'
+import { Store } from 'interfaces/store'
```
Dergash, line 2: Should use relative import (https://github.com/microsoft/TypeScript/issues/15479)

src/hooks/useDrillDownUrl.ts
```diff
@@ -0,0 +1,28 @@
...
+import {WidgetFieldBase} from '../interfaces/widget'
+import {buildBcUrl} from '../utils/strings'
+
+/**
+ * @param widgetName
+ * @param cursor
+ * @param fieldMeta
+ * @description Хук для возможности переопределения/отключения ссылки дриллдауна при вычислении ссылки в данных
```
Dergash, line 10: Proper TSDoc required (meaning english and no empty description/arguments)

Dergash (changes requested):

Dergash (approved):

PR #545, opened 2020-12-30
Full hierarchy list layout fixes

Lourence94:
Added ability to set width of last column, fixed -Infinitypx value of maxDepth

### DIFF
--- a/src/components/FullHierarchyTable/FullHierarchyTable.tsx
+++ b/src/components/FullHierarchyTable/FullHierarchyTable.tsx
@@ -87,8 +87,7 @@ export const FullHierarchyTable: React.FunctionComponent<FullHierarchyTableAllPr
     const loading = props.loading
     const depthLevel = props.depth || 1
     const levelValues = props.data?.map(item => item.level)
-    const maxDepth = (levelValues && Math.max(...levelValues)) || 1
-    console.warn(props.data)
+    const maxDepth = (levelValues && levelValues?.length && Math.max(...levelValues)) || 1
     const textFilters = React.useMemo(
         () => props.bcFilters?.filter(filter => [FilterType.contains, FilterType.equals].includes(filter.type)),
         [props.bcFilters]
--- a/src/utils/hierarchy.ts
+++ b/src/utils/hierarchy.ts
@@ -16,7 +16,7 @@
  */
 
 import { RowMetaField } from '../interfaces/rowMeta'
-import { WidgetField } from '../interfaces/widget'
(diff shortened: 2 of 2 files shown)


src/components/FullHierarchyTable/FullHierarchyTable.tsx
```diff
@@ -85,8 +85,8 @@ export const FullHierarchyTable: React.FunctionComponent<FullHierarchyTableAllPr
     const loading = props.loading
     const depthLevel = props.depth || 1
     const levelValues = props.data?.map(item => item.level)
-    const maxDepth = levelValues && Math.max(...levelValues) || 1
-    console.warn(props.data)
+    const maxDepth = levelValues && levelValues.length && Math.max(...levelValues) || 1
```
Dergash, line 88:
```tsx
levelValues?.length && Math.max(...levelValues) || 1
```

src/components/FullHierarchyTable/FullHierarchyTable.tsx
```diff
@@ -249,7 +249,7 @@ export const FullHierarchyTable: React.FunctionComponent<FullHierarchyTableAllPr
                     }
                 }))
         ]
-    }, [depthLevel, fields, props.meta.name, textFilters])
+    }, [depthLevel, fields, props.meta.name, textFilters, props.rowMetaFields])
```
Dergash, line 252: Memoization seems wrong on that one.
Aside of `props.rowMetaFields`, `maxDepth` is also used but not specified although it's obviously changes with data; `indentColumn` is also used but not specified and will need memoization of its own

src/utils/hierarchy.ts
```diff
@@ -52,16 +52,23 @@ export function getColumnWidth(
     const columnKey = showedFields?.indexOf(currentField)
     const currentColumnShift = currentField?.hierarchyShift
     const nextColumnShift = showedFields[columnKey + 1]?.hierarchyShift
+    // Ability to set last column width if one of previous columns have width = 0
+    const notFixedWidthColumn = showedFields.find(item => item.width === 0)
```
Dergash, line 56: Judging by the neighbour usage `showedFields` can and will be undefined at some point

Dergash: Also probably `noFixedWidthColumn`, as it's checking an existance of any fixed columns, not if specific one match this condition

Dergash (changes requested):

Dergash (approved):

Issue #530, opened 2020-12-14
Data and cursors for children business component not cleared when parent BC returns empty set


Dergash:
https://github.com/tesler-platform/tesler-ui/blob/87db3d98af93c339247692a3d87d73e5e6ccf588/src/epics/data.ts#L203

It appears that when data endpoint responds with empty data set we aren't clearing existing data and loading flag for descendant business components.

In practice it means two things:
1) `data` slice of the store will keep stale data for descendant business components
2) `loading` flag for descendant business components not cleared, and widget will be stuck showing spinner/skeleton.  
E.g. navigating to a view dispatches `selectView` action which sets `loading` flag for all present business components, and if parent BC will have no results all descendant widgets will be stuck.

Suggested fix
1)
```ts
// reducers/data.ts
case coreActions.bcFetchDataSuccess: {
    if (action.payload.data.length) {
        return state
    }
    const descendants = Object.values(store.screen.bo.bc)
    .filter(bc => bc.url.includes(`${action.payload.bcName}/:id`))
    .map(bc => bc.name)
    const newState = { ...state }
    descendants.forEach(bcName => {
        delete newState[bcName]
    })
    return newState
}
```

2)
```ts
// reducers/screen.ts
case coreActions.bcFetchDataSuccess: {
    if (action.payload.data.length) {
        return state
    }
    const newBoBc = { ...state.bo.bc }
    Object.values(store.screen.bo.bc)
    .filter(bc => bc.loading && bc.url.includes(`${action.payload.bcName}/:id`))
    .forEach(bc => {
        newBoBc[bc.name] = { ...state.bo.bc[bc.name], loading: false }
    })
    return {
        ...state,
        bo: {
            ...state.bo,
            bc: newBoBc
        }
    }
}
```

PR #567, opened 2021-01-18
If record is present, cursor shouldn't be reset (#566)

DRITE:
See #566 

### Addition
View selection shouldn't mark all its BCs as loading (see #530)
Otherwise we can reach next situation:
Let it be a view includes widget with BC `exampleBc` and type which is  included into `skipWidgetTypes`. If there is no one else widget with BC `exampleBc` then there is no data loading for `exampleBc` in fact. But action of view selecting has marked `exampleBc` as loading already. So if an application has only one such widget  then `exampleBc` become marked as loading permanently.

### DIFF
--- a/src/epics/data.ts
+++ b/src/epics/data.ts
@@ -130,6 +130,18 @@ export const bcFetchDataEpic: Epic = (action$, store) =>
                 .fetchBcData(state.screen.screenName, bcUrl, fetchParams, canceler.cancelToken)
                 .mergeMap(data => {
                     const newCursor = data.data[0]?.id
+                    const changeCurrentCursor = Observable.of<AnyAction>(
+                        $do.bcChangeCursors({
+                            cursorsMap: {
+                                [bcName]: data.data.some(i => i.id === cursor) ? cursor : newCursor
+                            },
+                            keepDelta: !!anyHierarchyWidget || (action.type === types.bcFetchDataRequest && action.payload.keepDelta)
+                        })
+                    )
+                    const changeCursors =
+                        action.type === types.bcFetchDataRequest && action.payload.depth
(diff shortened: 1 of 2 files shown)


src/epics/data.ts
```diff
@@ -130,6 +130,18 @@ export const bcFetchDataEpic: Epic = (action$, store) =>
                 .fetchBcData(state.screen.screenName, bcUrl, fetchParams, canceler.cancelToken)
                 .mergeMap(data => {
                     const newCursor = data.data[0]?.id
+                    const changeCurrentCursor = Observable.of<AnyAction>(
+                        $do.bcChangeCursors({
+                            cursorsMap: {
+                                [bcName]: data.data.length > 0 && cursor && data.data.some(i => i.id === cursor) ? cursor : newCursor
```
Dergash, line 136:
```
[bcName]: data.data?.some(i => i.id === cursor) ? cursor : newCursor
```
`?.some` already handles nullable in data response so no need to check for length explicitly; also if we have a record in our data with id `undefined` then we screwed too much for the check `cursor &&` to matter

src/reducers/screen.ts
```diff
@@ -295,6 +295,7 @@ export function screen(state = initialState, action: AnyAction): ScreenState {
                 newCache[bcName] = cursor
             })
             // Сбросить также курсоры у всех дочерних БК от запрошенных
+            // Also reset cursors of all children of requested BCs
```
Dergash, line 298: Russian comment should go

Dergash (changes requested):

src/epics/data.ts
DRITE: Condition is changed to `data.data.some(i => i.id === cursor)`

src/reducers/screen.ts
DRITE: Deleted

Dergash (approved):

Issue #593, opened 2021-02-09
Table Widget. Filter. Checkbox field type.


AndrewRAM33:
Found unexpected behavior of the filter in the table widget.

One of columns table widget has checkbox type, and table contains records. Make it possible to filter this column. Filter icon appears and popup displays checkbox - state necessary for filtering records. But when change value of filter checkbox indicator (without description, only checbox), there are no changes. The request to backend does not include filter field, as it is not in the redux store.

In the table widget, checkbox column filter is not applied.

(Custom dropdown filter layer does not used, only default behavior)

AndrewRAM33:
Present in Version 1.28.0 (Febraury 18, 2021)

PR #594, opened 2021-02-09
TableWidget filter for Checkbox field types

AndrewRAM33:
See issue #593

Changes have been made to the filter processing for checkbox column type.

Checkbox filter control incorrectly uses the event field that changes the indicator checkbox.
`e.target.value` instead `e.target.checked` (FilterField.tsx)

Need change the filter handler behavior. Filter value `false` for it should be subject-specific, not just empty, which means clear filter.

Please check correctness of my solution, and let me know if the fix is accepted so that I can update `tesler-ui` version in my app.

### DIFF
--- a/src/components/FilterPopup/FilterPopup.tsx
+++ b/src/components/FilterPopup/FilterPopup.tsx
@@ -70,7 +70,7 @@ export const FilterPopup: React.FC<FilterPopupProps> = props => {
             viewName,
             widgetName: widget.name
         }
-        if (!props.value) {
+        if (props.value === null || props.value === undefined) {
             dispatch($do.bcRemoveFilter({ bcName: widget.bcName, filter }))
         } else {
             dispatch($do.bcAddFilter({ bcName: widget.bcName, filter: newFilter }))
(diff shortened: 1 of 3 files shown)


src/components/FilterPopup/FilterPopup.tsx
```diff
@@ -70,7 +70,7 @@ export const FilterPopup: React.FC<FilterPopupProps> = props => {
             viewName,
             widgetName: widget.name
         }
-        if (!props.value) {
+        if (!(props.value || props.value === false)) {
```
Dergash, line 73: It probably would be better to directly checked for `null` and `undefined` values here as inverted checked is not readable without the context.
Otherwise seems ok.

Dergash (changes requested):

src/components/FilterPopup/FilterPopup.tsx
AndrewRAM33: Fixet to directly check for `null` and `undefined`

Dergash (approved):

PR #597, opened 2021-02-12
Operations on Hierarchy

ai-ba:
Waits for https://github.com/tesler-platform/tesler-schema/pull/1

Operations for Hierarchy.

### DIFF
--- a/package.json
+++ b/package.json
@@ -79,7 +79,7 @@
     "webpack-rxjs-externals": "1.1.0"
   },
   "dependencies": {
-    "@tesler-ui/schema": "0.2.0",
+    "@tesler-ui/schema": "0.3.0",
     "classnames": "2.2.6",
     "core-js": "3.1.4",
     "diff": "4.0.2",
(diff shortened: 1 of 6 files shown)


Dergash (commented):
Unfortunately it won't be merged, several factors contributed to this decision:

1) An existing solution for table widgets that served as a basis for this PR was never good in a first place: too many suspicious ref-based hacks working outside of usual React state management flow made this solution almost impossible to understand and maintain. The simple idea of showing `...` button at the end of hovered row should not rely on complex and obscure interaction of nine mutable references, each participating in hooks without properly being extracted to a dependency, which leads to implementation been very fragile.

2) An existing solution was not in fact reused but copy pasted, essentially doubling our resources to support both copies. It is obvious that existing solution was very complex, hence it should've been at least extracted to a separate component, properly documented and tested. We have at least two others hierarchy types and expanding this feature for them would meat to quadruple maintaining resources.

3) This feature is of limited usage - during alpha testing I was able to find only one widget where it was used, but the test coverage dropped 2.1% in this PR which does not seem acceptable.

Combined with other comments in this PR it seems that it will be easier to move forward with https://github.com/tesler-platform/tesler-ui/pull/638 instead.

PR #621, opened 2021-03-18
In case field type is pickList then ColumnFilter shows AssocListPopup for filtering

DRITE:
See #619

Additional: now condition of showing widget depends on `name` not `bcName`.

### DIFF
--- a/src/components/ColumnTitle/ColumnFilter.tsx
+++ b/src/components/ColumnTitle/ColumnFilter.tsx
@@ -12,6 +12,7 @@ import cn from 'classnames'
 import filterIcon from './filter-solid.svg'
 import FilterPopup from '../FilterPopup/FilterPopup'
 import FilterField from '../ui/FilterField/FilterField'
+import { PickListFieldMeta, WidgetTypes } from '@tesler-ui/schema'
 
 export interface ColumnFilterOwnProps {
     widgetName: string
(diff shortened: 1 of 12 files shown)


src/components/ColumnTitle/ColumnFilter.tsx
```diff
@@ -113,9 +121,14 @@ export const ColumnFilter: React.FC<ColumnFilterProps> = props => {
  */
 export function mapStateToProps(store: Store, ownProps: ColumnFilterOwnProps) {
     const widget = store.view.widgets.find(item => item.name === ownProps.widgetName)
+    const assocWidget = store.view.widgets.find(
```
Dergash: I think we should now move on to connecting components via [useSelector](https://react-redux.js.org/api/hooks#useselector) hook as its more backward compatibility friendly - `ColumnFilterProps` is a public interface and someone uses it the addition of mandatory `assocWidget` will crash the typings.
`<ColumnFilter />` is not realistically usable when not connected (internal use of `dispatch`), so connected props are also more of component internals then public API.

src/components/ColumnTitle/ColumnFilter.tsx
```diff
@@ -113,9 +121,14 @@ export const ColumnFilter: React.FC<ColumnFilterProps> = props => {
...
+        item =>
+            item.bcName === ((ownProps.widgetMeta as unknown) as PickListFieldMeta).popupBcName && item.type === WidgetTypes.AssocListPopup
```
Dergash: `as unknown` conversion is not needed here as `ownProps.widgetMeta` is of correct union type. The reason why this conversion was required in `handleVisibleChange` is that there variable was already explicitly converted to something else (`MultivalueFieldMeta`)

Dergash (approved):

PR #625, opened 2021-03-29
Extension point is added to PopupWidgetTypes

DRITE:
See #623

**Addition**: export of useDrillDownUrl hook


src/interfaces/widget.ts
```diff
@@ -192,15 +190,30 @@ export interface WidgetTextMeta extends WidgetMeta {
  */
 export type WidgetMetaAny = WidgetFormMeta | WidgetTableMeta | WidgetTextMeta | WidgetInfoMeta
 
+/**
+ * Component of custom widget
+ */
```
Dergash, line 195: We probably should deprecate this in favor of `React.ComponentType<any>` and remove in 2.0.0

src/components/View/View.tsx
```diff
@@ -71,3 +72,18 @@ function mapStateToProps(store: Store) {
...
+
+/**
+ * Add new values to `PopupWidgetTypes` from client application
+ *
+ * @param customWidgets
+ */
+function usePopupWidgetTypesExtension(customWidgets: Record<string, CustomWidgetDescriptor>) {
+    if (customWidgets) {
+        Object.entries(customWidgets).forEach(([widgetType, descriptor]) => {
+            if ((descriptor as CustomWidgetConfiguration).isPopup) {
```
Dergash, line 84: Use `isCustomWidget` typeguard:
https://github.com/tesler-platform/tesler-ui/blob/a5fdaca5e77958f9730b4a65aef4cbce4a820b9d/src/interfaces/widget.ts#L207

Dergash (changes requested):

src/interfaces/widget.ts
DRITE: Do you mean change 
```typescript
export type CustomWidget = ConnectedComponent<any, any> | FunctionComponent<any>
```
to
```typescript
export type CustomWidget = ConnectedComponent<any, any> | ComponentType<any>`
```
in case client has class components?

And I haven't got what are you talking about here:
> and remove in 2.0.0

src/components/View/View.tsx
DRITE: > Use `isCustomWidget` typeguard:
> https://github.com/tesler-platform/tesler-ui/blob/a5fdaca5e77958f9730b4a65aef4cbce4a820b9d/src/interfaces/widget.ts#L207

So in this case condition will be look like this:
```typescript
if (!isCustomWidget(descriptor) && descriptor.isPopup)
```
And you'll call it "Divine condition", aren't you? =)
Like here:
https://github.com/tesler-platform/tesler-ui/blob/master/src/components/Widget/Widget.tsx#L64

src/components/View/View.tsx
```diff
@@ -71,3 +72,18 @@ function mapStateToProps(store: Store) {
 const ConnectedView = connect(mapStateToProps)(View)
 
 export default ConnectedView
+
+/**
+ * Add new values to `PopupWidgetTypes` from client application
+ *
+ * @param customWidgets
+ */
+function usePopupWidgetTypesExtension(customWidgets: Record<string, CustomWidgetDescriptor>) {
```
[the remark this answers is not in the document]
DRITE: I've got it

src/components/View/View.tsx
```diff
@@ -71,3 +72,18 @@ function mapStateToProps(store: Store) {
 const ConnectedView = connect(mapStateToProps)(View)
 
 export default ConnectedView
+
+/**
+ * Add new values to `PopupWidgetTypes` from client application
+ *
+ * @param customWidgets
```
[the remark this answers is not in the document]
DRITE: OK

src/interfaces/widget.ts
Dergash: I meant:
```ts
/*
 * @deprecated TODO: Remove in 2.0.0
 */
export type CustomWidget = ComponentType<any>
```
The meaning of our `CustomWidget` type is simply "any react component declaration"; `ComponentType<any>` is exactly that as it's an alias for [class/FC union](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/0fa935e2b95688bc5e3bcb0290b1b3fd75323dec/types/react/index.d.ts#L78):
```ts
type ComponentType<P = {}> = ComponentClass<P> | FunctionComponent<P>;
```
`ConnectedComponent` is a bit trickier but overlaps enough with `ComponentType` to pass as one:
1) It's an [alias for NamedExoticComponent](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/d7ef36a683ef8b44ca1e478afed1cdb271a26671/types/react-redux/index.d.ts#L117) with some extensions
2) `NamedExoticComponent` is [an extension of ExoticComponent](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/d7ef36a683ef8b44ca1e478afed1cdb271a26671/types/react/index.d.ts#L357) with `displayName` prop
3) `ExoticComponent` is an interface for [function type of (props: P) => ReactElement | null](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/d7ef36a683ef8b44ca1e478afed1cdb271a26671/types/react/index.d.ts#L349) (with typeof) so it will pass as function component which is [the same](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/d7ef36a683ef8b44ca1e478afed1cdb271a26671/types/react/index.d.ts#L540) with a bunch of optional properties
  [derived](url) from `ComponentType`

DRITE: @Dergash Thank you

src/components/View/View.tsx
Dergash: Personally I'd vote for `Godlike` but in referenced link the confusion was about variable named `customWidget` when in fact it's a descriptor which may or may not be custom widget.

Here it's named `descriptor` so no such confusion applied.

Dergash (approved):

PR #634, opened 2021-04-13
ModalInvoke: Customization of button labels is added (#633)

DRITE:
see #633

### DIFF
--- a/src/components/ModalInvoke/ModalInvoke.tsx
+++ b/src/components/ModalInvoke/ModalInvoke.tsx
@@ -71,12 +71,16 @@ const ModalInvoke: React.FunctionComponent<ModalInvokeProps> = props => {
         }
     }
 
+    const okLabel = props.confirmOperation?.okText || t('Ok')
+    const cancelLabel = props.confirmOperation?.cancelText || t('Cancel')
+
     switch (props.confirmOperation.type) {
         case OperationPreInvokeType.info: {
             const modal = Modal.info({
                 className: styles.modal,
                 title: props.confirmOperation?.messageContent,
-                okText: t('Ok'),
+                okText: okLabel,
+                cancelText: cancelLabel,
                 onOk: () => {
                     props.onOk(props.bcName, props.operationType, props.widgetName, value || 'ok')
                     modal.destroy()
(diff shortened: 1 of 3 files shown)


src/components/ModalInvoke/ModalInvoke.tsx
```diff
@@ -71,12 +71,16 @@ const ModalInvoke: React.FunctionComponent<ModalInvokeProps> = props => {
         }
     }
 
+    const okLabel = props.confirmOperation.okText || t('Ok')
```
Dergash, line 74: Optional chaining probably required for `confirmOperation` (based on original line 78)

src/interfaces/operation.ts
```diff
@@ -205,20 +205,7 @@ export enum OperationPostInvokeConfirmType {
...
-    /**
-     * Title for modal
-     */
-    message: string
-    /**
-     * Additional text for modal
-     */
-    messageContent?: string
-}
+export type OperationPostInvokeConfirm = OperationModalInvokeConfirm
```
Dergash, line 208: I think we should reverse it: the key part is that we've received post invoke for confirmation, and then we can decide how to handle it. By default we handle it with modal window, but client application might be able to handle this type of post actions via popups, notifications or any other way.

So I think we should keep `OperationPostInvokeConfirm` and deprecate through alias `OperationModalInvokeConfirm`.

src/interfaces/operation.ts
```diff
@@ -229,13 +216,21 @@ export interface OperationModalInvokeConfirm {
      */
     type: OperationPostInvokeConfirmType | OperationPreInvokeType | string
     /**
-     * Title for modal
+     * Body text of a modal
      */
     message: string
     /**
-     * Additional text for modal
+     * Custom modal title
```
Dergash, line 223: Well that's misleading all right. Probably TODO for renaming will be appropriate.

Dergash (changes requested):

src/components/ModalInvoke/ModalInvoke.tsx
DRITE: Optional chaining is added

src/interfaces/operation.ts
DRITE: resolved

src/interfaces/operation.ts
DRITE: todo is added

Dergash (approved):

Issue #635, opened 2021-04-14
Need to add the ability to set background color for multivalueHover field type


artkzlv:
Need to add support for `backgroundColor` property in the MultivalueHover component.
Change the wrapping of multivalueHover field in tables: the \<Field /> should wrap \<MultivalueHover /> to be able to 
pass `commonProps` including `backgroundColor` to MultivalueHover.

Dergash:
Resolved in #635

PR #636, opened 2021-04-14
Need to add the ability to set background color for multivalueHover field type (#635)

artkzlv:
See #635

### DIFF
--- a/src/components/SameBcHierarchyTable/SameBcHierarchyTable.tsx
+++ b/src/components/SameBcHierarchyTable/SameBcHierarchyTable.tsx
@@ -5,9 +5,8 @@ import { Dispatch } from 'redux'
 import { $do } from '../../actions/actions'
 import { Store } from '../../interfaces/store'
 import Field from '../../components/Field/Field'
-import MultivalueHover from '../../components/ui/Multivalue/MultivalueHover'
 import { WidgetTableMeta, WidgetListField } from '../../interfaces/widget'
-import { DataItem, MultivalueSingleValue, PendingDataItem } from '../../interfaces/data'
+import { DataItem, PendingDataItem } from '../../interfaces/data'
 import { ColumnProps, TableRowSelection, TableEventListeners } from 'antd/lib/table'
 import { Route } from '../../interfaces/router'
 import { FieldType } from '../../interfaces/view'
@@ -36,8 +35,6 @@ export interface SameBcHierarchyTableProps extends SameBcHierarchyTableOwnProps 
     onDrillDown?: (widgetName: string, cursor: string, bcName: string, fieldKey: string) => void
     onExpand: (bcName: string, depth: number, cursor: string) => void
 }
-
-const emptyMultivalue: MultivalueSingleValue[] = []
 
 export const Exp: FunctionComponent = (props: any) => {
     if (!props.onExpand) {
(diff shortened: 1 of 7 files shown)


DRITE (dismissed):

src/components/widgets/TableWidget/TableWidget.tsx
```diff
@@ -364,31 +363,23 @@ export const TableWidget: FunctionComponent<TableWidgetProps> = props => {
...
                         return (
                             <div>
                                 <Field
                                     data={dataItem}
                                     bcName={props.meta.bcName}
                                     cursor={dataItem.id}
                                     widgetName={props.meta.name}
-                                    widgetFieldMeta={item}
+                                    widgetFieldMeta={
+                                        item.type === FieldType.multivalue ? { ...item, type: FieldType.multivalueHover } : item
```
Dergash, line 381: Will cause rerenders; should be memoized

src/components/SameBcHierarchyTable/SameBcHierarchyTable.tsx
```diff
@@ -158,17 +155,17 @@ export const SameBcHierarchyTable: FunctionComponent<SameBcHierarchyTableProps>
                 key: item.key,
                 dataIndex: item.key,
                 render: (text: string, dataItem: any) => {
-                    if (item.type === FieldType.multivalue) {
+                    if (item.type === FieldType.multifield || item.type === FieldType.multivalue) {
```
Dergash, line 158:
```ts
[FieldType.multifield, FieldType.multivalue].includes(item.type)
```

src/components/SameBcHierarchyTable/SameBcHierarchyTable.tsx
```diff
@@ -158,17 +155,17 @@ export const SameBcHierarchyTable: FunctionComponent<SameBcHierarchyTableProps>
...
                         return (
-                            <MultivalueHover
-                                data={(dataItem[item.key] || emptyMultivalue) as MultivalueSingleValue[]}
-                                displayedValue={item.displayedKey && dataItem[item.displayedKey]}
+                            <Field
+                                bcName={bcName}
+                                cursor={dataItem.id}
+                                widgetName={props.meta.name}
+                                widgetFieldMeta={item.type === FieldType.multivalue ? { ...item, type: FieldType.multivalueHover } : item}
```
Dergash, line 164: I don't think that arbitrary replacing one field type with another will lead us anywhere.

Seemingly correct way would be simply fix widgets descriptions themselves: find all table widgets, replace all `multivalue` fields with `multivalueHover` fields.   
If for some reason this can't be done it still should be handled on client application side:
```ts
// Custom component for multivalue field
const Multivalue = (props) => {
    const { widgetName } = props
    const widget = useSelector(store => store.view.widgets.find(item => item.name === widgetName))
    if (TableTypes.includes(widget.type)) {
        return <MultivalueHover {...props} />
    }
    return <Multivalue {...props} />
}
```
```ts
// AppLayout
<View
    customFields={{
        [FieldType.multivalue]: Multivalue
    }}
/>
```

Dergash (changes requested):

src/components/SameBcHierarchyTable/SameBcHierarchyTable.tsx
artkzlv: This method is chosen to keep the current behavior.
Yes, we can do this on the client side (with fix processing multivalue on tables in Tesler UI), but other client applications also expect to see multivalueHover instead of multivalue on tables.

src/components/widgets/TableWidget/TableWidget.tsx
```diff
@@ -349,6 +348,12 @@ export const TableWidget: FunctionComponent<TableWidgetProps> = props => {
         props.onSelectCell(recordId, props.meta.name, fieldKey)
     }
 
+    const fieldMeta = React.useMemo(
+        () => (item: WidgetListField) =>
+            item.type === FieldType.multivalue ? ({ ...item, type: FieldType.multivalueHover } as WidgetListField) : item,
+        []
```
Dergash, line 354: This memoization won't work; you declared a function `(item)  => ({ ...item, type })` and memoized it, and the function will always be the same, but you do not assigning a function to the prop - you are _calling_ a function and then  assigning its _result_ to the prop, and result of this function is new object on each call => each render => rerenders.

Try like this:
```ts
const fieldMeta = React.useMemo(() => {
return item.type === FieldType.multivalue ? ({ ...item, type: FieldType.multivalueHover  }) : item
, [item]}
```

Dergash (changes requested):

src/components/widgets/TableWidget/TableWidget.tsx
DRITE: @Dergash @artkzlv What do you think about removing field type conversion in 2.0.0 in favor of correctly built widget meta?
> I guess table widget shouldn't convert field type. Type of field should be resolved on meta data layer. If widget type is List then field type is multivalueHover. Otherwise multivalue.

Let's add corresponding TODO if there are no objections.

Dergash (approved):

src/components/widgets/TableWidget/TableWidget.tsx
Dergash: No objections

DRITE (approved):

Issue #612, opened 2021-03-01
Autosave infinity loop by `selectTableCellInit` action


AndrewRAM33:
This bug is reproducible on versions **tesler-ui** ^1.22.0

We have `view` containing `List` type widget as editable table:

```JSON
{
  "type": "List",
  "fields": [
    {
      "title": "Field_1",
      "key": "f_1",
      "type": "dictionary"
    },
    {
      "title": "Field_2",
      "key": "f_2",
      "type": "input"
    },
    {
      "title": "Field_3",
      "key": "f_3",
      "type": "checkbox"
    }
  ],
  "options": {
    "actionGroups": {
      "include": [
        "create",
        "save",
        "delete"
      ]
    }
  }
}
```

**Bug case:**
1. Create first record in the table by click create button, which executed `sendOperation.create` action
2. Fill required fields of the record, save by action `sendOperation.save`
3. Сreate another record by click button, it is called again `sendOperation.create` action and create record
4. When we want change value of field `dictionary` type in the last created record, by activating it from click on then, the application will freeze and crash

In developer tools, there will be memory overflow caused by an action looping in an autosave operation. The action will loop `selectTableCellInit`.

This case appears only after edit second added record and does not appear when edit first record. So it depends on the current cursor.

I tried to cause a case by first adding a record on one `BC`, then adding a record to another `BC`, and the loop also happened.

After check `AutosaveMiddleware` part of the code, I realized that the `selectTableCellInit` action is runing in the infinity loop because it has the `isNeedSaveCondition` condition after the cursor appears on the previously created record, I believe this is not correct.

```
autosaveMiddleware.ts

73    if (isNeedSaveCondition) {
74        return autosaveRoutine(action, { getState, dispatch }, next)
75    }
```

Please check autosave works. If this is actually wrong behavior, please fix it.

PR #637, opened 2021-04-15
`bcNewDataSuccess` resets selectedCell;  `operationRequiresAutosave` of `requiredFields` middleware ignores save operation (#612)

DRITE:
According the description #612 the reason of circular calling actions `sendOperation` with type `save` and `selectTableCellInit` is stale data in `selectedCell`. 

`selectedCell` isn't reseted after successful creating new table and it cheat on `autosaveMiddleware` [here](https://github.com/tesler-platform/tesler-ui/blob/master/src/middlewares/autosaveMiddleware.ts#L42) while processing of action `selectTableCellInit`.

Also `requiredFieldsMiddleware` reacts on `sendOperation` action with type `OperationTypeCrud.save`. This operation has `autoSaveBefore` flag. Since this operation is called on table widget `requiredFieldsMiddleware` calls `selectTableCellInit` action.

Hence we go to close loop.

To brake the loop we need to clear `selectedCell` by `bcNewDataSuccess` action . 
Also `requiredFieldsMiddleware` should ignore `autoSaveBefore` flag on save operation.

### DIFF
--- a/src/middlewares/requiredFieldsMiddleware.ts
+++ b/src/middlewares/requiredFieldsMiddleware.ts
@@ -11,9 +11,10 @@ import { openButtonWarningNotification } from '../utils/notifications'
 import i18n from 'i18next'
 import { DataItem, PendingDataItem } from '../interfaces/data'
 import { RowMetaField } from '../interfaces/rowMeta'
-import { WidgetField, WidgetFieldBlock, isWidgetFieldBlock, TableLikeWidgetTypes, WidgetTableMeta } from '../interfaces/widget'
-import { flattenOperations } from '../utils/operations'
+import { isWidgetFieldBlock, TableLikeWidgetTypes, WidgetField, WidgetFieldBlock, WidgetTableMeta } from '../interfaces/widget'
+import { checkOperationRole, flattenOperations } from '../utils/operations'
 import { PendingValidationFailsFormat } from '../interfaces/view'
+import { OperationTypeCrud } from '@tesler-ui/schema'
 
 const requiredFields = ({ getState, dispatch }: MiddlewareAPI<Dispatch<AnyAction>, CoreStore>) => (next: Dispatch) => (
     action: AnyAction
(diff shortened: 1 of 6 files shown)


src/middlewares/requiredFieldsMiddleware.ts
```diff
@@ -85,13 +86,15 @@ const requiredFields = ({ getState, dispatch }: MiddlewareAPI<Dispatch<AnyAction
...
-function operationRequiresAutosave(operationType: string, actions: Array<Operation | OperationGroup>) {
+export function operationRequiresAutosave(operationType: string, actions: Array<Operation | OperationGroup>) {
     let result = false
     if (!actions) {
         console.error('rowMeta is missing in the middle of "sendOperation" action')
         return result
     }
-    result = flattenOperations(actions).some(action => action.type === operationType && action.autoSaveBefore)
+    result = flattenOperations(actions).some(
+        action => action.type === operationType && action.autoSaveBefore && action.type !== OperationTypeCrud.save
```
Dergash, line 96: Ideally we should never check for `OperationTypeCrud.save` directly; a helper might be useful that check both `action.type` and `action.actionRole`.

src/middlewares/requiredFieldsMiddleware.ts
```diff
@@ -103,14 +106,14 @@ function operationRequiresAutosave(operationType: string, actions: Array<Operati
...
-function getRequiredFieldsMissing(record: DataItem, pendingChanges: PendingDataItem, fieldsMeta: RowMetaField[]) {
+export function getRequiredFieldsMissing(record: DataItem, pendingChanges: PendingDataItem, fieldsMeta: RowMetaField[]) {
     const result: PendingDataItem = {}
     fieldsMeta.forEach(field => {
         const value = record?.[field.key] as string
         const pendingValue = pendingChanges?.[field.key]
         const effectiveValue = pendingValue !== undefined ? pendingValue : value
         let falsyValue = false
-        if ([undefined, null, ''].includes(effectiveValue as any)) {
+        if ([undefined, null, ''].includes(effectiveValue as string)) {
```
Dergash, line 116: `any` conversion is more honest here as conversion to string relies heavily on `strictNullChecks` compiler option

Dergash (changes requested):

src/middlewares/requiredFieldsMiddleware.ts
DRITE: I get it

src/middlewares/requiredFieldsMiddleware.ts
DRITE: > Ideally we should never check for `OperationTypeCrud.save` directly; a helper might be useful that checks both `action.type` and `action.actionRole`.

It seems to be unnecessary to create another one utility like `matchOperationRole`.
Because exactly in this case we already iterate over `flattenOperations`.
I guess this condition could be resolved as followed here:
```typescript
result = flattenOperations(actions).some(
        action =>
            action.type === operationType &&
            action.autoSaveBefore &&
            action.type !== OperationTypeCrud.save &&
            action?.actionRole !== OperationTypeCrud.save
     
    )
```

DRITE: checkOperationRole helper is added

Dergash (approved):

PR #638, opened 2021-04-22
extract `<RowOperationsButton />` from `<TableWidget />`; reuse `<RowOperationsButton />` on `<HierarchyTable />`

Dergash:
Alternate approach to #597: instead of copying code to `<HierarchyTable />`, row operations button is extracted to separate component and reused.

### `<RowOperationsButton />`

* Reusable button with ref instance exposing `setRow` method to dynamically move button to the hovered row
* On click fetches row meta

### `<RowOperationsMenu />`

* Reusable menu with record-scoped operations
* On click sends operation

### useRowMenu

* Handles interactions between `<RowOperationsButton />` and `<Table />`: returns a ref for common parent, a ref for row operations button instance and `onMouseEnter`/`onMouseLeave` callback for `<Table onRow />` property

## useRowMenuInstance

* Extends `<RowOperationsButton />` instance with `setRow` method to programmatically move button to hovered row (called automatically from `onMouseEnter`/`onMouseLeave` callbacks)
* Extends `<RowOperationsButton />` instance with reference to its DOM node


src/components/RowOperations/RowOperationsButton.tsx
```diff
@@ -0,0 +1,114 @@
...
+ * {@link ActionPayloadTypes.bcSelectRecord | bcSelectRecord} and shows {@link RowOperationsMenu | RowOperationsMenu}
+ *
+ * {@link RowOperationsButtonInstance.setRow | setRow} can be used to dynamically place the button
+ * next to hovered row.
+ *
+ * @param props
+ * @param ref - Assigned reference will receive component instance
+ */
+export const RowOperationsButton: React.ForwardRefRenderFunction<RowOperationsButtonInstance, RowOperationsButtonProps> = (
+    { meta, parent, ...props }: RowOperationsButtonProps,
```
DRITE, line 50: I don't see any other props except `meta`, `parent`, and `bcName`.
May it be replaced with ```{ meta, parent, bcName }: RowOperationsButtonProps```?

DRITE (changes requested):

src/components/RowOperations/RowOperationsButton.tsx
Dergash: This will create naming collision as bcName inside the component is calculated from props.bcName and meta.bcName

DRITE: `const {bcName: propsBcName}: RowOperationsButtonProps`?
But it doesn't matter here. I've got it.

src/hooks/useRowMenu.ts
```diff
@@ -0,0 +1,105 @@
...
+                     * To position in the middle:
+                     *
+                     * rowTop - tableTop + height / 2 - buttonSize / 2
+                     */
+                    containerRef.current.style.top = `${rowTop - tableTop + 17}px`
+                    onRowHover(record.id)
+                }
+            }
+        }),
+        [parent?.current, containerRef?.current, skip, onRowHover, onRowLeave]
```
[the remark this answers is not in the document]
Dergash: Apparently this is intentional

1) Why destructuring for 1 variable?
2) Extracting ref from dependency array outside of the `useImperativeHandle` [will break](https://github.com/tesler-platform/tesler-ui/actions/runs/785677799) the behavior. Refs are mutable, hence they do not cause rerender when `current` is changed, hence `useRowMenu` will not be called again when its parent changes refs `current`, hence both container and parent will be undefined and will never be initialized.
Current behavior works because it's a callback, and actual `current` value is accessed during user interaction. My understanding was that refs are never should be part of dependencies arrays at all (as `ref` itself is consistent, and `ref.current` not causes rerenders); extracting `current` to separate variable is not a correct way.

I suppose we'll need to do some reading about how eslint plugin works, if we should remove all refs from all dependencies or there is a possibility for stale state and we should change our approach with their usage.

DRITE (approved):

Issue #661, opened 2021-07-13
Performing `postAction` of type `confirmation` during saving process


DRITE:
There was a request from customers to implement ability of using `confirm` `postAction` during `sendOperation` of type  `save`.

DRITE:
The request from customers was canceled as they can use `preInvoke` instead.

PR #662, opened 2021-07-13
postAction confirmations is precessed during `sendOperation` `save`

DRITE:
See #661


DRITE:
Issue is canceled. See #661

PR #685, opened 2021-08-20
Date filters should use locally selected date even if actual timestamp falls off to another day due to the timezone

Dergash:


src/reducers/screen.ts
```diff
@@ -298,11 +300,15 @@ export function screen(state = initialState, action: AnyAction): ScreenState {
         }
         case types.bcAddFilter: {
             const { bcName, filter } = action.payload
+            const isDate = store.view.widgets
+                .find(item => item.name === action.payload.widgetName)
+                ?.fields.find((item: WidgetField) => item.type === FieldType.date && item.key === action.payload.filter.fieldName)
```
DRITE, line 305: Let's use `filter` from already destructured `action.payload` object. 
I mean `item.key === filter.fieldName`

src/reducers/screen.ts
```diff
@@ -298,11 +300,15 @@ export function screen(state = initialState, action: AnyAction): ScreenState {
         }
         case types.bcAddFilter: {
             const { bcName, filter } = action.payload
+            const isDate = store.view.widgets
+                .find(item => item.name === action.payload.widgetName)
```
DRITE, line 304: What about to place `widgetName` to destructuring of `action.payload`?

DRITE (changes requested):

src/reducers/screen.ts
```diff
@@ -408,6 +414,26 @@ export function screen(state = initialState, action: AnyAction): ScreenState {
...
+ *
+ * @param timestamp Zero timezoned timestamp, e.g. `2021-08-19T22:07:16.192Z`
+ */
+function correctDateFilter(timestamp: string) {
+    const originalDate = new Date(timestamp)
+    const newDate = new Date(
+        originalDate.getFullYear(),
+        originalDate.getMonth(),
+        originalDate.getDate(),
+        -(originalDate.getTimezoneOffset() / 60)
```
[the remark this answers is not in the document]
Dergash: [getTimezoneOffset](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/getTimezoneOffset#negative_values_and_positive_values) returns signed number which is the difference between UTC and local timezone in minutes.

The point of the minus is to negate the effect of local timezone. If the user is left of Greenwich then getTimezoneOffset is positive number.

For example, the user is in Greenland (UTC-3), left of Greenwich and it's _august 19_ and 23:00 on his clock when he picks a date from the calendar. So the ISO is `2021-08-19T23:00:00.000-03:00` for that moment, but we store timestamps in zero timezone, so the actual value of the filter is `2021-08-20T02:00:00.000Z` - Greenwich is already in _august 20_. But the user didn't care about Greenwich, he selected august 19 on his calendar, so our way of storing date leads us to sending next day to Tesler API.

So, we take `2021-08-20T02:00:00.000Z` from the filter and substract timezone offset (3 hours), which gives us `2021-08-19T23:00:00.000Z` - essentially for the purposes of filtration if we store dates in Z-time, then user also should select it in Z-time instead of his local time. Calendar works in local so we negate timezone.

Right of Greenwich is the same except `getTimezoneOffset` gives negative values, which combined with our own minus sign gives us plus, so there is addition of local timezone and negative timezoneoffset to get back into Z-time.

src/reducers/screen.ts
```diff
@@ -408,6 +414,26 @@ export function screen(state = initialState, action: AnyAction): ScreenState {
     }
 }
 
+/**
+ * Date fields are filtred by the way of selecting of local timestamp in calendar, transforming it into zero timezoned timestamp
```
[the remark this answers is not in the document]
Dergash: Fixed in https://github.com/tesler-platform/tesler-ui/pull/688/files

src/reducers/screen.ts
Dergash: Fixed in https://github.com/tesler-platform/tesler-ui/pull/688/files

src/reducers/screen.ts
Dergash: Fixed in https://github.com/tesler-platform/tesler-ui/pull/688/files

src/reducers/screen.ts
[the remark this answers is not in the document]
DRITE: i've got it

Issue #691, opened 2021-08-25
Hide elements in AssocListPopup hierarchy


AndrewRAM33:
**extension request:**

Need to enable hiding items in `AssocListPopup` widgets with hierarchy. It is necessary in order to prevent the user from associating specific elements of the hierarchy.

Dergash:
What kind of hiding you have in mind, id-based, like an array of ids to exclude in widget json?

Generally I'd say it makes more sense to exclude records by filtering them out in Tesler API service.

AndrewRAM33:
Solved by a custom implementation of the multivalue field. Thanks

PR #692, opened 2021-08-26
Hide elements in AssocListPopup hierarchy

AndrewRAM33:
see #691

**Extension:** Ability to hide records from the user association in `AssocListPopup`.

Records are excluded by setting of `_hidden` flag in the widget data. At the same time, `TAG` is not displayed, and association `_assotiate` remains active.

`MVG` field will also display hidden records and correctly delete their association.

_example data:_
``` js
data: [
	...
	{
 		businessProcess: "test bp",
		id: 1,
		level: 1,
		name: "test bp",
		number: "1",
		parentId: "0",
		vstamp: 2,
		_associate: true,
		_hidden: true,	// dont't show in popup hierarchy and TAG
	}
	...
]
```

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -2,6 +2,7 @@ on:
   push:
     branches:
       - master
+  workflow_dispatch:
 
 name: Build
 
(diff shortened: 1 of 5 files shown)


AndrewRAM33:
Solved by a custom implementation of the `multivalue field`. Thanks

PR #703, opened 2021-09-08
`ViewNavigationWidget` and `NavigationTabsWidget` are added

DRITE:
See #702

### NavigationTabsWidget
The purpose of the widget is to display navigation tabs depends on position in view's metadata. In other words this widget should be included in widgets stream.

### ViewNavigationWidget
The widget displays the same component as `NavigationTabsWidget`.
But this widget should be used in case need to exclude navigation  from widgets stream.For instance, client could decide to fix it on top of the screen.

### DIFF
--- a/src/components/ui/NavigationTabs/NavigationTabs.tsx
+++ b/src/components/ui/NavigationTabs/NavigationTabs.tsx
@@ -1,0 +1,45 @@
+/*
+ * TESLER-UI
+ * Copyright (C) 2018-2020 Tesler Contributors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import React from 'react'
+import { Tabs } from 'antd'
+import styles from './NavigationTabs.less'
+import { useViewTabs } from '../../../hooks'
+import { historyObj } from '../../../reducers/router'
+
+interface NavigationTabsProps {
+    navigationLevel: number
+}
+
+function NavigationTabs({ navigationLevel }: NavigationTabsProps) {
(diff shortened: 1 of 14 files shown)


src/components/ui/NavigationTabs/NavigationTabs.tsx
```diff
@@ -0,0 +1,45 @@
...
+function NavigationTabs({ navigationLevel }: NavigationTabsProps) {
+    const tabs = useViewTabs(navigationLevel)
+    const handleChange = (key: string) => {
+        historyObj.push(key)
+    }
+
+    return (
+        <nav className={styles.container}>
+            <Tabs activeKey={tabs?.find(item => item.selected)?.url} tabBarGutter={24} size="large" onChange={handleChange}>
+                {tabs.map(item => (
```
Dergash, line 37: No null check when there is one on previous line

src/components/widgets/NavigationTabsWidget/NavigationTabsWidget.tsx
```diff
@@ -0,0 +1,36 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import React from 'react'
+import NavigationTabs from '../../ui/NavigationTabs/NavigationTabs'
+import { NavigationWidgetMeta } from '../../../interfaces/widget'
+
+export interface NavigationTabsWidgetProps {
+    meta: NavigationWidgetMeta
```
Dergash, line 23: Why the entire meta object is a prop when only `navigationLevel` field is actually used?

src/hooks/useViewTabs.ts
```diff
@@ -40,7 +40,7 @@ function mapStateToProps(store: Store): UseViewTabsState {
  * @param depth 1 for top level navigation; 2, 3, 4 for SecondLevelMenu, ThirdLevelMenu and FourthLevelMenu
  * @category Hooks
  */
-export function useViewTabs(depth: NavigationLevel) {
+export function useViewTabs(depth: NavigationLevel | number) {
```
Dergash, line 43: There is no point in this union: the resulting type will be extended to `number`. If all numbers now supported, `NavigationLevel` should be removed

Dergash (changes requested):

src/components/ui/NavigationTabs/NavigationTabs.tsx
DRITE: fixed

src/hooks/useViewTabs.ts
DRITE: removed

Dergash (approved):

Issue #710, opened 2021-09-28
Replace date/datetime `equals` filter with a pair of `greaterOrEqualThan` and `lessOrEqualThan`


Dergash:
Currently when user applies a filter to `date`/`dateTime` field it's an `equals` filter which is send to Tesler API endpoint and later split into two different search specifications on backend, replacing time part with beginning of the day and the end of the day.

It is required to have a [range picker](https://ant.design/components/date-picker/#components-date-picker-demo-range-picker) as a filter control instead and instead of `equals` filter send a pair of `greaterOrEqualThan` and `lessOrEqualThan` to Tesler API endpoint.

PR #723, opened 2021-10-22
changed component DatePicker to RangePicker

achabalov:
See #710 
filterField changed component DatePicker to RangePicker
reducer screen.ts bcAddFilter added date check. 321 lines look bad and bcRemoveFilter added check on field fieldName


achabalov:
close. See #733

PR #736, opened 2022-05-12
Moved Header from TableWidget

romadim:


src/components/widgets/TableWidget/components/Header.tsx
```diff
@@ -0,0 +1,88 @@
...
+    const { t } = useTranslation()
+    const [filterGroupName, setFilterGroupName] = React.useState(null)
+
+    const { filtersExist, limitBySelf, filterGroups, cursor } = useSelector((store: Store) => {
+        const bc = store.screen.bo.bc[bcName]
+        const bcCursor = bc?.cursor
+        const filters = store.screen.filters[bcName]
+
+        return {
+            filtersExist: filters?.length,
```
Dergash, line 26: You might want to convert into boolean here
1) Your effect only cares for false value, but will fire on each change in length
2) In render you checking it like ` {filtersExist && ... }`, if filters array is empty the number "0" will be shown in UI

romadim: Thank you. I fixed it.

PR #748, opened 2022-09-26
fixed old cursor in checkShowCondition

artemego:
checkShowCondition was using an old cursor value


DRITE:
@artemego  provide a detailed description, please

**N.B.** *For future PRs:* We have [contribution guide](https://github.com/tesler-platform/tesler-ui/blob/master/CONTRIBUTING.md). Please check it out. According to the guide, you must follow the name formats `feat/issue-number` or `fix/issue-number`. This means that if there is no issue created you have to create the issue yourself.

artemego:
For list widgets with showCondition set, infinite loading occurs after data requests. This seems to happen because the checkShowConditon function does not use the changed cursor. For example, if you switch the page of the list widget, then checkShowCondition will have a cursor from the previous page, so the condition that the widget is hidden will work and the data and rowMeta requests will be canceled

